PR #4, opened 2019-04-16
Updates the PETSc version for Travis-CI

bishtgautam:
PETSc version for Travis-CI is updated from v3.10.4 to v3.11.1


nocollier:
Oh, ok I will be more careful when I commit on a different PETSc version. Thanks!

PR #5, opened 2019-04-16
Adds support for the MPFA-O for 2D problem

### LINKED ISSUE #6: Extend MPFA-O method to support 2D meshes in parallel
Presently, MPFA-O method only works for serial runs.

bishtgautam:
Support for Multi Point Flux Approximation (MPFA) - O method is added.
Presently, MPFA-O methods works with 2D meshes for serial and parallel runs.
Additionally, a Fortran and C code for the MPFA-O method is added.

Fixes #6


bishtgautam:
Finally Travis-CI has a successful build.

bishtgautam:
@nocollier, This is ready for a review and merge.

demo/mpfao.c
```diff
@@ -0,0 +1,289 @@
...
+}
+
+void Pressure3D(double *x,double *f) {
+  (*f)  = PetscSqr(x[0])*PetscSqr(-x[0]+1);
+  (*f) += PetscSqr(x[1])*PetscSqr(-x[1]+1);
+  (*f) += PetscSqr(x[2])*PetscSqr(-x[2]+1);
+}
+
+void Velocity3D(double *x,double *v) {
+  double K[9]; Permeability(x,K);
```
nocollier:
```suggestion
  double K[9]; Permeability3D(x,K);
```

demo/mpfao.c
```diff
@@ -0,0 +1,289 @@
...
+  v[0] = -K[0]*(-2*x[0]+1) - K[1]*(-2*x[1]+1) - K[2]*(-2*x[2]+1);
+  v[1] = -K[3]*(-2*x[0]+1) - K[4]*(-2*x[1]+1) - K[5]*(-2*x[2]+1);
+  v[2] = -K[6]*(-2*x[0]+1) - K[7]*(-2*x[1]+1) - K[8]*(-2*x[2]+1);
+}
+
+void Forcing3D(double *x,double *f) {
+  double K[9]; Permeability(x,K);
```
nocollier:
```suggestion
  double K[9]; Permeability3D(x,K);
```

demo/mpfao.c
```diff
@@ -0,0 +1,289 @@
+#include "tdycore.h"
```
nocollier: I am wondering if we really need this file? As far as I can see you set it up to do what the steady.c does. Can we just use steady.c and then change the method with `-tdy_method {wy|bdm|tpf|mpfa_o}`?

demo/mpfao.c
```diff
@@ -0,0 +1,289 @@
...
+      break;
+    case 4:
+      ierr = TDySetPermeabilityFunction(tdy,PermeabilitySineFunction,NULL); CHKERRQ(ierr);
+      ierr = TDySetForcingFunction2(tdy,ForcingSineFunction,NULL); CHKERRQ(ierr);
+      ierr = TDySetDirichletValueFunction(tdy,PressureSineFunction,NULL); CHKERRQ(ierr);
+      ierr = TDySetDirichletFluxFunction(tdy,VelocitySineFunction,NULL); CHKERRQ(ierr);
+      break;
+    }
+  } else {
+    SETERRQ(PETSC_COMM_WORLD,PETSC_ERR_USER,"MPFA-O does not support 3D problems");
```
nocollier: This is not needed as you have a check in TDyMFPAOInitialize if dim != 2

demo/mpfaof90.F90
```diff
@@ -0,0 +1,225 @@
+module f90module
```
nocollier: This is great and I hope goes a long way to help people see that this could work trivially in fortran codes

src/tdycoremesh.c
```diff
@@ -0,0 +1,1230 @@
...
+}
+
+/* -------------------------------------------------------------------------- */
+PetscErrorCode AllocateMemoryForASubcell(
+  TDy_subcell    *subcell,
+  TDySubcellType subcell_type) {
+
+  PetscFunctionBegin;
+
+  PetscInt num_nu_vectors;
```
nocollier: Because these get set in the select statement, the compiler complains that they might not be initialized. Can we initially set them to 0 to squelch this warning?

src/tdycoremesh.c
```diff
@@ -0,0 +1,1230 @@
...
+  PetscInt       num_vertices,
+  PetscInt       num_edges,
+  PetscInt       num_neighbors,
+  TDySubcellType subcell_type) {
+
+  PetscFunctionBegin;
+
+  PetscErrorCode ierr;
+  PetscInt       isubcell;
+  PetscInt       num_subcells;
```
nocollier:
```suggestion
  PetscInt       num_subcells=0;
```

src/tdycoremesh.c
```diff
@@ -0,0 +1,1230 @@
...
+  TDy_edge       *edges;
+  PetscInt       icell, iedge;
+  PetscInt       ncells, nedges;
+  PetscReal      x,y;
+  PetscReal      theta[8];
+  PetscInt       idx[8], ids[8], subcell_ids[8];
+  PetscInt       tmp_subcell_ids[4];
+  PetscInt       tmp_cell_ids[4], tmp_edge_ids[4], tmp_ncells, tmp_nedges;
+  PetscInt       count;
+  PetscInt       start_idx;
```
nocollier:
```suggestion
  PetscInt       start_idx=-1;
```
I am not sure what to initialize this to, but because it first gets set in a conditional statement, the compiler complains.

src/tdympfao.c
```diff
@@ -0,0 +1,1601 @@
...
+  for (iedge=0; iedge<mesh->num_edges; iedge++) {
+
+    if (!edges[iedge].is_internal) { // Is it a boundary edge?
+
+      // Determine the cell ID for the boundary edge
+      if (edges[iedge].cell_ids[0] != -1) icell_1 = edges[iedge].cell_ids[0];
+      else                                icell_2 = edges[iedge].cell_ids[1];
+
+      // Is the cell locally owned?
+      if (cells[icell_1].is_local) edges[iedge].is_local = PETSC_TRUE;
```
nocollier: `icell_1` is uninitialized if the above conditional is false

src/tdympfao.c
```diff
@@ -0,0 +1,1601 @@
...
+/* -------------------------------------------------------------------------- */
+
+PetscErrorCode TDyMPFAORecoverVelocity(TDy tdy, Vec U) {
+
+  DM             dm;
+  TDy_mesh       *mesh;
+  TDy_vertex     *vertices, *vertex;
+  TDy_edge       *edges, *edge;
+  PetscInt       ivertex, icell, isubcell;
+  PetscInt       icol, row, col, vertex_id, iedge;
```
nocollier: same deal, `row` is set inside a conditional, initialize it to something.

bishtgautam:
@nocollier, Thanks for the review and I will make the suggested changes. 

Regarding the following question
> can we not have mpfao.c in the demo directory and just make it work with 'steady -tdy_method mpfao' ?

In the long run, I agree that mpfao.c can be simply replaced with steady.c. In the short run, there are differences between how MPFA and WY/BDN/TPF are currently using user defined functions to set permeability, forcing, dirichlet and dirichlet flux. It is not completely obvious to me which is the correct approach, so I'm using mpfao.c as a playground to test things.

src/tdympfao.c
bishtgautam: The else condition should be setting `icell_1` instead of `icell_2`. It is now fixed in 7f9be5c1f4d42e8b4bafe7f50e9bd6c595a975ee

src/tdycoremesh.c
bishtgautam: I instead add an error message under the `default` case within the `switch(subcell_type)` https://github.com/TDycores-Project/TDycore/pull/5/commits/e7441c7e94d399ab7f76cee3617631efa9a1ae3a#diff-fcf52e8a4105ab2e05e6a4f90c5dfbd9R33

src/tdycoremesh.c
bishtgautam: Similarly, I added an error message under the `default` case within the `switch(subcell_type)` https://github.com/TDycores-Project/TDycore/pull/5/commits/e7441c7e94d399ab7f76cee3617631efa9a1ae3a#diff-fcf52e8a4105ab2e05e6a4f90c5dfbd9R87

src/tdycoremesh.c
bishtgautam: Fixed by https://github.com/TDycores-Project/TDycore/pull/5/commits/e7441c7e94d399ab7f76cee3617631efa9a1ae3a#diff-fcf52e8a4105ab2e05e6a4f90c5dfbd9R555

src/tdympfao.c
bishtgautam: Fixed via https://github.com/TDycores-Project/TDycore/pull/5/commits/e7441c7e94d399ab7f76cee3617631efa9a1ae3a#diff-d55e8fd8865fbd3ac7d2d195600dcc2fR1266

PR #12, opened 2019-06-03
Add Fortran support for WY method

bishtgautam:
Updates the WY method to also support `tdy->ops` along with the `SpatialFunction`. 
This change allows one to reuse the current Fortran interface. A regression test for
F90+WY is added.

Additionally, the cell ID (or key) in the regression file is fixed to be the natural cell ID.


src/tdywy.c
```diff
@@ -725,7 +755,12 @@ PetscReal TDyWYPressureNorm(TDy tdy,Vec U) {
     ierr = DMPlexGetPointGlobal(dm,c,&gref,&junk); CHKERRQ(ierr);
     if(gref<0) continue;
     ierr = PetscSectionGetOffset(sec,c,&offset); CHKERRQ(ierr);
-    tdy->dirichlet(&(tdy->X[c*dim]),&p);
+    if (tdy->dirichlet){
+      tdy->dirichlet(&(tdy->X[c*dim]),&p);
+    }
+    if (tdy->ops->computedirichletvalue) {
```
nocollier: Is there a reason why this is in here twice? There are a few places where you are moving from XX to ops->XX where things are repeated. Are you thinking to just add in the new method and phase out the old?

bishtgautam: I'm hoping that we move to the `tdy->ops->XYZ()`. What do you think? If that is okay, I can make additional commits in this PR to swap out all old methods with new methods.

nocollier: I am ok with moving to ops, this is needed for the fortran interface correct?

bishtgautam: Yes, those changes are for the fortran interface. I will make the necessary changes in this PR.

bishtgautam:
@nocollier, It is ready for your review. I just realized that I'm passing `TDy tdy` argument but am never using it. Should I remove it now? Or, should we fix it later?

nocollier:
I cannot think of a case for where these functions need to know about the TDy, that would be if we were interpolating at a point in the domain. And we *have* added a context to pass in any specific parameters we need to. Perhaps it is better to remove dependence? I am not opposed to later either.

bishtgautam:
You are correct and I will clean up the PR.

bishtgautam:
@nocollier, After struggling for a while I realize that I cannot get rid of first argument (i.e. `TDy tdy`) because we are using `PetscObject[S/G]etFortranCallback()` whose first argument is the object itself. Let's merge the PR as is and we can clean it up if needed in the future.

PR #15, opened 2019-06-13
Adds regression tests for BDM and TPF

bishtgautam:
4 steady tests are added for each of the two methods


nocollier:
Looks good.

PR #16, opened 2019-06-14
Fixes the TPF and BDM tests

bishtgautam:
Adds the missing `-tdy_method <tpf|bdm>` in the test configuration
file


bishtgautam:
@nocollier Don't merge it as it seems the BDM fails for few tests.

bishtgautam:
The PR is good to go now. We are expected to have >90% code coverage!!!

nocollier:
Nice man, I wonder if the missing coverage are debugging functions that I only use when implementing things.

bishtgautam:
There are a bunch of functions in [tdycoremesh.c](https://codecov.io/gh/TDycores-Project/TDycore/src/master/src/tdycoremesh.c) for outputting info about the mesh for MPFA-O. Those functions aren't currently tested. Need to add a command line option so one can output the mesh and then update a test to include that option.

PR #17, opened 2019-07-04
Adds an option to output mesh connectivity

bishtgautam:
By adding `-tdy_output_mesh` to command line, the mesh attributes and connectivity
information is outputted. Additionally, a regression test is updated to test the new option.

### DIFF
--- a/include/private/tdymeshimpl.h
+++ b/include/private/tdymeshimpl.h
@@ -133,6 +133,7 @@ struct _TDy_mesh {
 
 };
 
+PETSC_EXTERN PetscErrorCode OutputMesh(TDy);
 PETSC_EXTERN PetscErrorCode BuildTwoDimMesh(TDy);
 PETSC_EXTERN PetscErrorCode AllocateMemoryForMesh(DM,TDy_mesh*);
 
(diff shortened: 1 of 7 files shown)


include/private/tdymeshimpl.h
```diff
@@ -133,6 +133,7 @@ struct _TDy_mesh {
 
 };
 
+PETSC_EXTERN PetscErrorCode OutputMesh(TDy);
```
nocollier: Should these calls (and 137-8) be prefixed with TDy to avoid possible name collisions?

nocollier:
Do I understand correctly that this provides an option to output the plex into some format that you know how to plot? And you already wrote the code that does it, this just exposes an option?

include/private/tdymeshimpl.h
bishtgautam: This is a good catch. But, I now notice that there are many places in the lib that I have not added a prefix `TDy`. How about if I fix the lib by adding `TDy` prefix in a separate PR?

nocollier:
Sounds fine to me, I wasn't paying close enough attention either.

bishtgautam:
>Do I understand correctly that this provides an option to output the plex into some format that you know how to plot? And you already wrote the code that does it, this just exposes an option?

You are exactly correct. In the future, we can add an option to dump out the data not only in binary but also in hdf5, provided PETSc was installed with HDF5 support.

PR #18, opened 2019-07-08
Extends MPFA-O for 3D problems

### LINKED ISSUE #7: Extend MPFA-O method to support 3D meshes
Presently MPFA-O methods only works for 2D grids in serial

bishtgautam:
The MPFA-O implementation is extended for 3D grids. A regression test is added
for a 3D problem.

Fixes #7


bishtgautam:
Let me rebase this branch off the latest master so as to avoid any conflicts

bishtgautam:
@nocollier, I completely agree with your assessment above. In addition to the two issues that you have raised, I feel the current algorithm needs to be simplified as I myself keep getting confused. I believe boiling down the algorithm into a pseudo-code would be a start that we should tackle later this month.

PR #21, opened 2019-07-11
Modifications to improve SPE performance

nocollier:
Removes entries in the system matrix when a diagonal permeability is used on a regular grid. 
This would result in fewer non-zero entries and reduce the cost of MatVec products. 
Additionally, adds some accessors for TDy data such as the dimension such that we need
not include private headers in application code.


src/tdywy.c
```diff
@@ -554,17 +554,20 @@ PetscErrorCode TDyWYComputeSystem(TDy tdy,Mat K,Vec F) {
...
     /* C and D are in column major, but C is always symmetric and D is
        a vector so it should not matter. */
+    PetscReal maxC = 0;
+    for(c=0; c<nB*nB; c++) { maxC = PetscMax(maxC,PetscAbsReal(C[c])); }
     for(c=0; c<nB; c++) {
       ierr = DMPlexGetPointGlobal(dm,Bmap[c],&gStart,&junk); CHKERRQ(ierr);
       if(gStart < 0) continue;
       ierr = VecSetValue(F,gStart,D[c],ADD_VALUES); CHKERRQ(ierr);
       for(q=0; q<nB; q++) {
+	if (PetscAbsReal(C[q*nB+c])<(1e-12*maxC)) continue;
```
bishtgautam: How about defining `1e-12` as a variable (e.g. `PetscReal toleranceFactor = 1e-12;`)?

PR #28, opened 2019-10-01
Updates for the transient problem

nocollier:
This pull request includes functionality needed to run transient problems and includes:

* VanGenuchten saturation functions (ported from PFLOTRAN)
* BrooksCorey relative permeability (ported from PFLOTRAN)
* bug fixes in the assigning of the porosity


bishtgautam:
@nocollier, Could you cherry-pick 566ba752c72fbc9019a3b4d107b27deff73f09d4 in your branch or update the `.travis/install-petsc.sh` to use gitlab instead of bitbucket for PETSc?

src/tdywy.c
```diff
@@ -918,14 +926,16 @@ PetscErrorCode TDyWYResidual(TS ts,PetscReal t,Vec U,Vec U_t,Vec R,void *ctx) {
...
+        div += sign*tdy->vel[nv*(f-fStart)+j]*wgt*tdy->V[f];
       }
     }
 
-    r[c] = tdy->porosity[c-cStart]*tdy->dS_dP[c-cStart]*dp_dt[c] + div;
-
+    r[c] = tdy->porosity[c-cStart]*tdy->dS_dP[c-cStart]*dp_dt[c] + div - tdy->Flocal[c-cStart];
+    //PetscPrintf(PETSC_COMM_WORLD,"R[%2d] = %+e %+e %+e = %+e\n",
+    // 	c,tdy->porosity[c-cStart]*tdy->dS_dP[c-cStart]*dp_dt[c],
+    // 		div,tdy->Flocal[c-cStart],r[c]);
```
bishtgautam: Consider deleting these lines.

nocollier:
I wasn't sure how to run the tests manually. I am fairly certain there will be differences that will require gold file updates.

src/tdywy.c
nocollier: We aren't 'out of the woods' yet and I might still need to refer to them. Cleaning in a future commit.

bishtgautam:
> I wasn't sure how to run the tests manually.

What??? :p

* From the home directory, run all tests via `make rtest`, or
* `cd regression_tests; make test-transient`

src/tdywy.c
bishtgautam: Okay.

nocollier:
Ok, I see now. I poked around makefiles and just didn't see the particular target. We will need to discuss this a bit.

src/tdywy.c
```diff
@@ -918,14 +926,16 @@ PetscErrorCode TDyWYResidual(TS ts,PetscReal t,Vec U,Vec U_t,Vec R,void *ctx) {
       sign = PetscSign(TDyADotBMinusC(&(tdy->N[dim*f]),&(tdy->X[dim*f]),
                                       &(tdy->X[dim*c]),dim));
       for(j=0; j<nv; j++) {
-        div += sign*tdy->vel[dim*(f-fStart)+j]*wgt*tdy->V[f];
+        div += sign*tdy->vel[nv*(f-fStart)+j]*wgt*tdy->V[f];
```
bishtgautam: Why is `wgt(= 1/nv)` used in computation of `div`? Also, is `rho` missing in the computation of `div` and `r[c]` ([below](https://github.com/TDycores-Project/TDycore/pull/28/files#diff-55c021d81f73d9728a8c237b533b9688R933))?

nocollier: Yes, I have taken rho = constant for now and therefore cancels out of both sides of the equation. I don't want to introduce more complexity before I figure out our current problems. However, we need that and some sort of pressure dependence (Glenn suggested a linear model in PFLOTRAN) so we can solve when fully saturated. Current tests simply don't hit that state.

nocollier: The weight bit appears to be a mistake as well, nice catch.

PR #40, opened 2020-04-07
Satkarra/th mpfao

satkarra:
Adds the following new capabilities:

- added TH for MPFAO discretization
- modified tdyregression to accommodate 2 dofs per cell
- added demo - transient_th_mpfao under the transient directory
- added *vtk output in the demo (eventually we should think of adding this to the TDycore library).
- created regression using this demo


satkarra:
Gautam,

Can you take a look at it now? The tdyregression re-factor is causing issues. So, I've scratched it for now. I'll re-visit with a fresh code base. If you are okay with the rest, please merge. 

Satish

bishtgautam:
@satkarra Thanks for the fix. I will review this PR tomorrow.

PR #59, opened 2020-07-04
Enable testing from top-level makefile

jedbrown:
You can now run

    make -j4 test

to run the full test suite in parallel, rebuilding library and demos as needed. Or

    make -j4 test-steady

to run any single test, also rebuilding (in parallel) anything that's needed. Use `make print-ALLTESTS` to list all options.

We could remove pretty much all the other makefiles, but I've left them intact for now. We could also place the compiled demos into `PETSC_ARCH`, which I've also skipped to avoid disrupting a workflow that moves into the demo directory.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -1,0 +1,11 @@
+# Default target
+all :
+	+@$(MAKE) -f makefile --no-print-directory $@
+
+ifeq ($(firstword $(sort 4.1.99 $(MAKE_VERSION))),4.1.99)
+include gmakefile
+endif
+
+# For any target that doesn't exist in gmakefile, use the legacy makefile (which has the logging features)
+% :
+	+@$(MAKE) -f makefile --no-print-directory $@
--- a/demo/SPE10/makefile
+++ b/demo/SPE10/makefile
@@ -15,6 +15,6 @@ TDYCORE_DIR ?= $(topdir)
 include $(TDYCORE_DIR)/lib/tdycore/conf/variables
 include $(TDYCORE_DIR)/lib/tdycore/conf/rules
 
-steady: steady.o chkopts
+steady: steady.o
 	$(CLINKER) -o $@ $< $(TDYCORE_LIB) $(LIBS)
 	$(RM) -f $<
--- a/demo/mpfao/makefile
+++ b/demo/mpfao/makefile
@@ -23,7 +23,7 @@ include $(TDYCORE_DIR)/lib/tdycore/conf/rules
 build:
 	@cd ../; make
 
-mpfao: mpfao.o chkopts
+mpfao: mpfao.o
 	$(CLINKER) -o $@ $< $(TDYCORE_LIB) $(LIBS)
 	$(RM) -f $<
 
--- a/demo/steady/makefile
+++ b/demo/steady/makefile
@@ -23,6 +23,6 @@ include $(TDYCORE_DIR)/lib/tdycore/conf/rules
 build:
 	@cd ../../; make
 
-steady: steady.o chkopts
+steady: steady.o
 	$(CLINKER) -o $@ $< $(TDYCORE_LIB) $(LIBS)
 	$(RM) -f $<
--- a/demo/steadyblock/makefile
+++ b/demo/steadyblock/makefile
@@ -23,6 +23,6 @@ include $(TDYCORE_DIR)/lib/tdycore/conf/rules
 build:
 	@cd ../../; make
 
-steadyblock: steadyblock.o chkopts
+steadyblock: steadyblock.o
 	$(CLINKER) -o $@ $< $(TDYCORE_LIB) $(LIBS)
 	$(RM) -f $<
--- a/demo/steadyf90/makefile
+++ b/demo/steadyf90/makefile
@@ -23,6 +23,6 @@ include $(TDYCORE_DIR)/lib/tdycore/conf/rules
 build:
 	@cd ../; make
 
-steadyf90: steadyf90.o chkopts
+steadyf90: steadyf90.o
 	$(FLINKER) -o $@ $< $(TDYCORE_LIB) $(LIBS)
 	$(RM) -f $<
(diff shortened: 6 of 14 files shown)


bishtgautam:
@jedbrown, 

- When I run `make demo` twice in `demo/steadyf90`, I get the following error:

```
>make clean
>make
/Users/bish218/projects/petsc/petsc/darwin-gcc8-with-blaslapack-openmpi/bin/mpif90 -c -Wall -ffree-line-length-0 -Wno-unused-dummy-argument -g   -I/Users/bish218/projects/petsc/petsc/include -I/Users/bish218/projects/petsc/petsc/darwin-gcc8-with-blaslapack-openmpi/include -I/opt/X11/include -I/usr/local/include    -I/Users/bish218/projects/tdycore/TDycore-master/darwin-gcc8-with-blaslapack-openmpi/include -I/Users/bish218/projects/tdycore/TDycore-master/include    -o steadyf90.o steadyf90.F90
Warning: chkopts target is deprecated and can be removed from user makefiles
/Users/bish218/projects/petsc/petsc/darwin-gcc8-with-blaslapack-openmpi/bin/mpif90 -Wl,-multiply_defined,suppress -Wl,-multiply_defined -Wl,suppress -Wl,-commons,use_dylibs -Wl,-search_paths_first -Wl,-no_compact_unwind -Wl,-multiply_defined,suppress -Wl,-multiply_defined -Wl,suppress -Wl,-commons,use_dylibs -Wl,-search_paths_first -Wl,-no_compact_unwind  -Wall -ffree-line-length-0 -Wno-unused-dummy-argument -g  -o steadyf90 steadyf90.o -Wl,-rpath,/Users/bish218/projects/tdycore/TDycore-master/darwin-gcc8-with-blaslapack-openmpi/lib -L/Users/bish218/projects/tdycore/TDycore-master/darwin-gcc8-with-blaslapack-openmpi/lib -ltdycore -Wl,-rpath,/Users/bish218/projects/petsc/petsc/darwin-gcc8-with-blaslapack-openmpi/lib -L/Users/bish218/projects/petsc/petsc/darwin-gcc8-with-blaslapack-openmpi/lib -Wl,-rpath,/Users/bish218/projects/petsc/petsc/darwin-gcc8-with-blaslapack-openmpi/lib -L/Users/bish218/projects/petsc/petsc/darwin-gcc8-with-blaslapack-openmpi/lib -Wl,-rpath,/opt/X11/lib -L/opt/X11/lib -Wl,-rpath,/opt/local/lib/gcc8/gcc/x86_64-apple-darwin18/8.3.0 -L/opt/local/lib/gcc8/gcc/x86_64-apple-darwin18/8.3.0 -Wl,-rpath,/opt/local/lib/gcc8 -L/opt/local/lib/gcc8 -lpetsc -lHYPRE -lcmumps -ldmumps -lsmumps -lzmumps -lmumps_common -lpord -lscalapack -lp4est -lsc -llapack -lblas -lexodus -lnetcdf -lpnetcdf -lhdf5hl_fortran -lhdf5_fortran -lhdf5_hl -lhdf5 -lparmetis -lmetis -ltriangle -lz -lX11 -lstdc++ -ldl -lmpi_usempif08 -lmpi_usempi_ignore_tkr -lmpi_mpifh -lmpi -lgfortran -lquadmath -lm -lstdc++ -ldl
/bin/rm -f -f steadyf90.o
> make
m2c    -o steadyf90.o steadyf90.mod
make: m2c: No such file or directory
make: *** [steadyf90.o] Error 1
```

- I believe in your first comment you meant to say `make -j4 rtest` instead of `make -j4 test`

- I'm all for moving the compiled demos into `PETSC_ARCH`, though it doesn't have to be in this PR.

jedbrown:
I really do mean `make test` from the top level, though that requires GNU Make 4.2 or higher. With older make, it'd be `make -f gmakefile test`. The `rtest` target is recursive make in a way that can't (reasonably) have accurate dependencies.

bishtgautam:
`make -f gmakefile test` does work.

jedbrown:
Great. PETSc has this somewhat complicated makefile system so it can support non-GNU make (that runs GNU make internally) and for better logging of output (`make.log`) that can be emailed. The `GNUmakefile` has a version guard because (at least in PETSc), there was a bug with this usage in older versions. (I don't recall the details, but could recreate and test.) I don't know what the requirements are here for building demos, but it can be reliable and small code to drop the legacy walks.

bishtgautam:
Regarding the error when building `demo/steadyf90` twice, it seems that the change of module name to [steadyf90](https://github.com/TDycores-Project/TDycore/pull/59/files#diff-bbec90d0c339a23eac039e2cb800825eL1) and the filename being `steadyf90.F90` is somehow messing up the make.

jedbrown:
Thanks. I updated the branch.

I'd be happy to move the executables, but would prefer doing it while also getting rid of `demo/*/makefile` to avoid needing to support/document both modalities.

jedbrown:
Also, what are our requirements on version of `make`? Do we need to support running `bsdmake foo` to invoke (via `$(OMAKE)`) GNU Make, or can we expect users to invoke GNU Make themselves? If the latter, I assume it'll still need to work back to 3.82.

bishtgautam:
@jedbrown , I'm not that familiar with `make` to answer your question and will leave it up to you and others to decide which version of make to support.

Regarding getting rid of `demo/*/makefile`. My current workflow is (i) make changes in `src/` and compile the lib, (2) compile one of the demos and run a simulation. If we get rid of `demo/*/makefile`, will be able to still compile a single demo.

jedbrown:
@bishtgautam Yeah, you would `make test-mpfao` or whatever, all from the top level. You wouldn't have to compile the library separately (it's done as part of that command when needed). I'd do `M-x compile` in Emacs and set `make -C ~/src/libceed test-mpfao` and then hit `M-x recompile` (bound to a key) after each change (wherever that may be). Changing directories and running different commands is an expensive unnecessary context switch, IMO.

bishtgautam:
@nocollier, Do you have any comments on this PR?

PR #60, opened 2020-07-04
WIP: Q2 mixed

jedbrown:
1. [x] Sketch residual using PetscDS interfaces
2. [ ] Use PETSc `src/snes/tutorials/ex69.c` (which solves Stokes) as a model to compute the coupled residual and Jacobian. Verify accuracy in the `steady` test (which set up in this branch, though the exact numbers will likely change).
3. [ ] Implement support to assemble the divergence and mass operators separately, preferably using `MatGetLocalSubMatrix`.
4. [ ] Replace assembly of the mass matrix (the expensive part) in the previous line with a `MatShell` using libCEED.
5. [ ] Use a "SIMPLE" approximation of the Schur complement `B diag(M)^{-1} B^T` with a PCFieldSplit preconditioner and algebraic multigrid on the Schur complement.

@rezgarshakeri


knepley:
Don't look at SNES ex62, since a lot of things are very old. I will update it sometime soon. However, ex69 has nonlinear Stokes and is much more modern.

jedbrown:
Thanks, Matt. @rezgarshakeri ex69.c has thousands of lines of generated code for evaluating manufactured solutions (SolCx, SolKx). I'd skip to `main` and check `SetupDiscretization`, which is similar to what I've sketched here.

rezgarshakeri:
> We can't add it this way because it hides everything in that subdirectory. You could `demo/steady/steady` to hide the executable only, though IMO, we should move those into a build directory.

Sorry, I didn't know that. I'll fix it.

rezgarshakeri:
> Could you please amend this commit to remove the PDF and binary files?

My bad. Sure.

bishtgautam:
@jedbrown @rezgarshakeri Do you want to keep this PR open?

rezgarshakeri:
> @jedbrown @rezgarshakeri Do you want to keep this PR open?

No, I think we can close it.

PR #63, opened 2020-07-20
Glenn/refactor dm creation

ghammond86:


bishtgautam:
@ghammond86, The vertex perturbations are problem-specific and we should not code them up in the TDycore itself. I guess the `SetVertexPerturbationFunction(PerturbDMInteriorVertices)` isn't taking in `TDy` argument because  `TDyCreate(&tdy)` creates the DM and one cannot perturb the vertices after DM is already created and distributed. But how about removing the `DMPlexDistribute` and moving it to `TDySetDiscretizationMethod` as:

```
  - TDycreate(tdy)
     - TDyCreateDM(DM) // Create a box mesh or read exodus file, but don't call DMPlexDistribute

  - TDySetVertexPerturbationFunction(tdy, <user-function>)

  - TDySetDiscretizationMethod()
    - if (tdy->ops->computemeshperturbation) (*tdy->ops->computemeshperturbation)(tdy,...)
    - DMPlexDistribute()
    - rest of the code:
```

ghammond86:
@bishtgautam TDySetDiscretizationMethod() is called within TDySetFromOptions() when the discretization method is specified from the command line, and TDySetFromOptions() is called after TDySetDiscretizationMethod() in the demos.  Is it safe to re-distribute a "distributed" DM?

bishtgautam:
I had not thought about the scenario you mentioned and I need to find another solution.

ghammond86:
Changes were revised in #65 to avoid refactoring demo tests.

PR #78, opened 2020-09-14
Neumann boundary condition and two-point flux in MPFA-O

bishtgautam:
- Command line option,`-tdy_mpfao_boundary_condition`, can now be used to specify
Dirichlet or Neumann boundary condition for MPFA-O method (eg. `MPFAO_DIRICHLET_BC` `MPFAO_NEUMANN_BC`).
Currently only zero-flux Neumann boundary condition is implemented.

- The MPFA-O method can reduced to a two-point flux via `-tdy_mpfao_gmatrix_method MPFAO_GMATRIX_TPF`.

- Tests are added for new capabilities.


bishtgautam:
@jeff-cohere I refactored internals of the MPFA-O method significantly. The higher-level logging events that you are working on shouldn't be impacted, but lower level logging events may have a conflict.

jeff-cohere:
Thanks for the heads-up. I'll put my profiling changes up as a PR so you can see what I've been up to.

PR #79, opened 2020-09-14
Added a registry for timers, and macros for creating/starting/stopping them.

jeff-cohere:
This PR adds some macros that can be used to do more comprehensive performance profiling without needing to actively manage the lifetimes of PETSc's `PetscLogEvent` objects.

Briefly:
* If you include the `tdytimers.h` header, you get access to a number of macros there. The most useful ones do function-level profiling: `TDY_START_FUNCTION_TIMER` and `TDY_STOP_FUNCTION_TIMER` can be placed at the beginning and the end of a function to create a managed timer that shows up in profiling reports.
* You can generate a profiling report with the `-tdytimers` input option. This generates a `tdycore_profile.log` with timing information.
* There are initialization and break-down functions for timers as well. It might be helpful to create a `TDyInit` function to set up the entire environment (calling `PetscInitialize`, too) in which these could be handled.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,5 @@
 *xcode/*
+*.d
+*.o
+*.vtk
+*.DIR
--- a/include/tdytimers.h
+++ b/include/tdytimers.h
@@ -15,7 +15,7 @@ PETSC_EXTERN PetscClassId TDY_CLASSID;
 // values. PETSc ships with khash, so let's appropriate it.
 #include <petsc/private/kernels/khash.h>
 KHASH_MAP_INIT_STR(TDY_TIMER_MAP, PetscLogEvent)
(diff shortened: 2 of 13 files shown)


bishtgautam:
An example of profiling information generated by this PR is at https://gist.github.com/bishtgautam/56cffdd86927376dc83fb1a4f5680ee2. `TDy` stuff shows under PETSc's main stage.

@rtmills I have added you as a reviewer as may have some feedback.

jeff-cohere:
So how should we handle stages? The timers so far might be in the main stage because they're all located outside of any `PetscLogStagePush` <-> `PetscLogStagePop` blocks.

bishtgautam (approved):

PR #80, opened 2020-09-16
Rosie/add exodus output

leorosie:


src/tdydm.c
```diff
@@ -114,6 +114,13 @@ PetscErrorCode TDyCreateDM(DM *dm) {
     ierr = DMPlexCreateBoxMesh(PETSC_COMM_WORLD, dim, PETSC_FALSE, faces,
                                lower, upper, NULL, PETSC_TRUE, dm);
     CHKERRQ(ierr);
+
+    int num_cell_in_set;
+    num_cell_in_set = Nx * Ny * Nz;
```
ghammond86, line 119: Nx, Ny, Nz are the global values.  What happens in parallel when each process sets the label value for all cells?

ghammond86: The solution is to wrap these line within:

   PetscInt rank;
    ierr = MPI_Comm_rank(PetscObjectComm((PetscObject)*dm),&rank);
    if (!rank) {
      ...
    }

so that only process 0 sets the labels.  This is what was causing master to fail.  When I wrap the setting of the label, this works for v3.13 and the tip of master.

ghammond86 (approved):

ghammond86:
@bishtgautam Do you want us to rebase and squash here too?

bishtgautam:
Yes, if that isn't too much work. Also, could one of the tests be updated to output data?

ghammond86:
Superseded by PR #89

PR #83, opened 2020-09-20
Added TDyInit, TDyFinalize, and profiling stage registry.

### LINKED ISSUE #82: Create TDyInit() (or equivalent) for initializing PETSc, timers, log stages, and other subsystems
Right now, we have several demos/drivers that all perform their own initialization of PETSc and whatever else is needed. As a step toward "library-izing" TDycore, I'd like to introduce an initialization function that serves the same role for TDycore as `PetscInitialize` does for PETSc.

The most immediate reason for this is that we now have timers, and we'll soon define log event stages to be used by those timers. I think it's less confusing for everyone if we initialize the timers subsystem and register those stages as soon as possible so that no one tries to time anything before that's done.

Suppose we call this initialization function `TDyInit()`. Maybe it accepts `argc` and `argv` arguments. Then all the demos and drivers can replace their calls to `PetscInitialize` with `TDyInit`. We can also add a `TDyFinalize` function, or we could just have it register a cleanup function to be called on exit.

## Log Event Stages

Because this is such a trivial reorganization, why not also take this opportunity to introduce some simple named stages? Here are my suggestions:

* "TDycore Setup", for creating meshes, setting up initial conditions (and whatever other bookkeeping we need)
* "TDycore Stepping", for timestepping
* "TDycore I/O", for checkpointing, restarting, generating visualizations.

jeff-cohere:
This PR adds initialization functions for the TDycore library. It also adds a registry for named profiling stages similar to that for timers. This is an attempt to normalize the way we do performance profiling so we don't all have to separately keep track of our own `PetscLogEvent` and `PetscLogStage` objects/definitions.

For more details, see the [Initializing the TDycore Library](https://github.com/TDycores-Project/TDycore/wiki/Initializing-the-TDycore-Library) and [Times and Profiling](https://github.com/TDycores-Project/TDycore/wiki/Timers-and-Profiling) pages.

Closes #82


knepley:
@jeff-cohere Hi Jeff. I edited the Times and Profiling page to note that any number of logging stages can be used. This stuff looks good.

bishtgautam (approved):
Looks great.

bishtgautam:
@jeff-cohere How about updating one of the tests (e.g. `demo/richards/richards.cfg`) to include `-tdy_timers`?

jeff-cohere:
Good idea.

PR #89, opened 2020-09-28
refactored TDyIO to include exodus

ghammond86:
This is the rebased and squashed version of Rosie's updates

### DIFF
--- a/.travis/install-petsc.sh
+++ b/.travis/install-petsc.sh
@@ -2,7 +2,7 @@
 
 git clone https://gitlab.com/petsc/petsc.git
 
-PETSC_GIT_HASH=1a6d72e33c
+PETSC_GIT_HASH=a826417c4d
 DEBUG=1
 
 cd petsc
@@ -31,5 +31,3 @@ export PETSC_DIR=$PWD
 
 
 make all
-#make test
-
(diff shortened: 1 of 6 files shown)


leorosie (commented):
Did we want to add the wrapper for the cell labels?

bishtgautam:
Is one of a test using the new I/O capability?

bishtgautam (approved):

ghammond86:
> Did we want to add the wrapper for the cell labels?

Yes.

ghammond86:
> Is one of a test using the new I/O capability?

Since our testing harness lacks the ability to compare .exo formats, I don't know that a test helps other than to prove that the simulation ran to completion.  What do you think?

bishtgautam:
With the new test we will make sure that new PRs don't break current I/O capability. Also, the will provide an example of how to use the new I/O.

ghammond86:
> > Did we want to add the wrapper for the cell labels?
> 
> Yes.

I will add the wrapper.

ghammond86:
> With the new test we will make sure that new PRs don't break current I/O capability. Also, the will provide an example of how to use the new I/O.

I looked into this further.  The Exodus output is hardwired to the file 'out.exo' in the richards_driver executable.  There are no command line arguments to test.  I recommend that we add some command line arguments for setting the format, output filename and perhaps variable names and frequency.  At that point, add a few tests.  @leorosie can work on this.

Issue #24, opened 2019-07-18
Add documentation for the library


bishtgautam:

bishtgautam:
@jeff-cohere commentted regarding documentation on [PR-100](https://github.com/TDycores-Project/TDycore/pull/100#issuecomment-719025564) and I'm moving that discussion here.

@jeff-cohere, I think adding documentation would be useful, but I don't have experience enough to advocate for one or the other approach.

- PETSc has its own very sophisticated system of creating the doc from the comments on the code. Not sure if we should try to use that framework.
- I checked [libCEED](https://github.com/CEED/libCEED) and it has user+dev guide within `src/doc`, which is hosted on [readthedocs](https://libceed.readthedocs.io/en/latest/). Could not figure out if doxygen is being used by libCEED. (cc: @jedbrown). 

Do others have suggestions/experience regarding how should we go about creating documentation for the library?

jeff-cohere:
Doxygen is a decent out-of-the-box approach. It was written primarily for use with C++, but I know some tricks to get it to work well with C. If PETSc's approach can be applied to TDycore, that'd be another good option.

nocollier:
It appears that libCEED does use Doxygen at least to generate some
intermediate xml.

https://github.com/CEED/libCEED/blob/main/doc/README.md

When I think about documentation that I use vs that I don't, I am usually
googling things. I look a lot at python docs like
numpy/scipy/netCDF/xarray/pandas but more often than not I get there via a
google search. So I would advocate whatever we choose being online so it
gets indexed.

I also spend a lot of time looking around at the PETSc docs which I tend to
access directly via their website because I don't know what function I
should call. I love that they have a lot of their API on one page that I
can just CTRL-F to find functions that might work for me. For example, I
used CTRL-F "PetscSpaceSet" yesterday to figure out how I should create a
space for a test I was running. However, I also use the PETSc documentation
a lot with "highlight function" + "ESC" + "." inside emacs (uses etags).
This takes me right to the source code and I can read the docs inline.

Nate

jedbrown:
Yep, libCEED uses Doxygen-style comments and a Sphinx plugin called Breathe to create its [API pages](https://libceed.readthedocs.io/en/latest/api/CeedBasis/). Some editors (e.g., VS Code) have smart rendering for such comments. I would _not_ recommend using Sowing (the man page tooling used by PETSc and MPICH) in new projects.

[...]

PR #100, opened 2020-10-13
Adds support to break the 3D domain

bishtgautam:
- A connected region PETSc binary file can be specified via `-tdy_connected_region`.
  The file contains region ID for each cell in the 3D mesh.
- No flow is enforced between cells with different region IDs.
- Presently this new feature is only supported for `-tdy_mpfao_gmatrix_method MPFAO_GMATRIX_TPF`
- A new test is added.


demo/transient/transient_snes_mpfaof90.F90
```diff
@@ -185,6 +186,21 @@ program main
     endif
   endif
 
+  call DMGetDimension(dm, dim, ierr);
```
jeff-cohere, line 189: Is this new code block related to the parallelism problem you found a week or so back? Is it something that we could pull into a function for convenience?

demo/transient/transient_snes_mpfaof90.cfg
```diff
@@ -1,6 +1,7 @@
 [suites]
```
jeff-cohere, line 1: At some point, it might be nice to add comments describing the problems/initial conditions described by the parameters in these files.

src/mesh/tdycoremesh.c
```diff
@@ -489,6 +492,49 @@ PetscErrorCode SaveMeshGeometricAttributes(TDy tdy) {
...
+
+      // Save the region ids
+      TDy_region *region = &mesh->region_connected;
+
+      ierr = VecGetLocalSize(local, &size_local);
+      region->num_cells = size_local/num_fields;
+      ierr = TDyAllocate_IntegerArray_1D(&region->id,region->num_cells); CHKERRQ(ierr);
+
+      ierr = VecGetArray(local, &p); CHKERRQ(ierr);
+      for (ii = 0; ii < size_local/num_fields; ++ii) region->id[ii] = p[ii*num_fields];
```
jeff-cohere, line 529: You can use `region->num_cells` here instead of `size_local/num_fields` to make it a bit more obvious what's going on.

jeff-cohere:
Quick question: what kind of feedback are you looking for? Much of the coding style is congruent with how things are done in PETSc, but some things could perhaps be better encapsulated if you envision them being useful in more than one context.

Also (and this is a topic for after the ESMD meeting), I think it would be good to start documenting code as we go, instead of assuming that we'll have the time and energy to do it afterward (when we've started to forget the details...). Depending on how much we want to take a "library" approach for TDycore, we might even want to generate reference documentation with, say, [Doxygen](https://www.doxygen.nl/index.html).

jeff-cohere:
By the way, let me know if this PR is time critical to merge, and I'll approve. The approach looks sound and is easy enough to understand.

demo/transient/transient_snes_mpfaof90.F90
bishtgautam: This code block isn't related to the parallel problem. 

If a demo is creating a DM and using that to create TDycore, then this code block allows mapping of the local cell id to natural cell id. If TDycore is created without a DM, then TDycore does this automatically.

bishtgautam:
This PR doesn't need to be merged in master before the ESMD PI meeting. 

>Quick question: what kind of feedback are you looking for? Much of the coding style is congruent with how things are done in PETSc, but some things could perhaps be better encapsulated if you envision them being useful in more than one context.

Let's implement better encapsulation. If you think this PR is a good place to then let's do it.

demo/transient/transient_snes_mpfaof90.F90
jeff-cohere: Is this something we'd like to wrap in a function to make available for other demos?

src/mesh/tdycoremesh.c
```diff
@@ -369,6 +369,9 @@ PetscErrorCode TDyAllocateMemoryForMesh(TDy tdy) {
   mesh->num_subcells = cNum*num_subcells;
   ierr = AllocateMemoryForSubcells(cNum, num_subcells, subcell_type, &mesh->subcells); CHKERRQ(ierr);
 
+  TDy_region *region = &mesh->region_connected;
```
jeff-cohere, line 372: It might be nice, for this and for the other structs here (edges, faces, etc) to provide functions for creating and manipulating their state, and for destroying it when we're done. In the case of the `TDy_region` (which we could rename to `TDyRegion` to be more consistent with the other types and with the PETSc way of naming things:

```
// Creates a new empty region to which mesh cells can be added
PetscError  TDyRegionCreate(TDyRegion* region) {
  region->num_cells = 0;
  region->id = NULL;
  return 0;
}

// Adds the cells with the given IDs to the region.
PetscErrorCode TDyRegionAddCells(TDyRegion* region,
                                                         PetscInt numCells,
                                                         PetscInt cellIDs[numCells]) {
   region->num_cells += numCells;
   PetscError err = PetscRealloc(region->num_cells, &region->id);
   if (err != 0) return err;
   err = PetscMemcpy(region->id, cellIDs);
   return err;
}

// Destroys a region, deallocating all resources.
void TDyRegionDestroy(TDyRegion* region) {
  if (region->id != NULL) {
    TDyDeallocate_IntegerArray_1D(region->id);
  }
}
```

And so on and so forth. This isolates the code specific to a region and allows you to understand where you're doing allocations and all that.

bishtgautam: @jeff-cohere  Thanks for this example. In the coming days, I will refactor this PR for regions using the above example.

bishtgautam:
@jeff-cohere, Following your example, I have added [include/private/tdyregionimpl.h](https://github.com/TDycores-Project/TDycore/pull/100/files#diff-7ab0fee4d0245d84fb8efaff89328baf112c37d8fd52e80862ef6ed6abcc9942) and [src/mesh/tdyregion.c](https://github.com/TDycores-Project/TDycore/pull/100/files#diff-6553f780922827b027540ba1bb6b234d34b6df3fb026dbaf87ce12f09b28b796). But, I'm not sure how to protect other TDycore internal source codes from modifying the `TDyRegion` struct i.e. how to disable setting of `region->id` instead of calling `TDyRegionAddCells()` on [tdycoremesh.c#L530](https://github.com/TDycores-Project/TDycore/pull/100/files#diff-2e8de59d7f94a447d9166144c1a228e05325d894e96df2c448c399bbeb52d867R530)?

jeff-cohere:
Hey @bishtgautam ,

When you use the world "protect" here, you're talking about encapsulation--the practice of hiding stuff from developers that you don't want them to think about. Object-oriented languages like C++ and Java have a lot of support for hiding these things built in. C doesn't really provide us with anything, so we have to do it ourselves.

I have mixed feelings about how much time to spend trying to protect developers from themselves--focusing too much on this can make everything feel clunky and hard to work with. I like hiding allocation/deallocation logic from scientific programmers, because most of them don't want to think about that stuff, and many of them mess things up when they do try to do it.

For accessing cell indexes within a region, though, sometimes we just need to provide some hints. In this case, changing `region->id` to something like `region->cells` or `region->cell_ids` might give everyone a better idea about how to use it (especially since `region->num_cells` is already named well).

Another way to prevent people from Doing Bad Things is to document things--either with comments/annotations in headers (using Doxygen, for example, and generating an API reference guide) or by describing the `TDyRegion` data type in the Wiki. In either case we can show some examples that illustrate how to add cells to a region, and how to loop over them to do region-related things.

What do you think about renaming the `TDyRegion` type's `id` field to `cell_ids` and adding some documentation?

bishtgautam:
>C doesn't really provide us with any such thing

That is what I thought was the case, but I wasn't sure if other devs had an approach to circumvent this deficiency of C. I tried to look at how PETSc was dealing with this issue but I wasn't able to figure it out. Developing our own encapsulation might not be worth it right now.

Should only the `<struct>.c` have functions to access/modify the data of the struct e.g. code change in [tdympfao3D_core.c](https://github.com/TDycores-Project/TDycore/pull/100/commits/a6b109ef68ee4f6e0afd3da3dc5e3b816818804f#diff-7b291e729681b9342f1e12b563a9577c86bd161f21e48b6489007901a606af0eL916)?

Moved the discussion about the documentation to #24.

jeff-cohere:
We don't have to consider C's lack of explicit encapsulation as a deficiency, per se--one can really go overboard in making a code impossible to use improperly (and to all of the C++ and Java evangelists who argue otherwise: I can break your code if I want to!). I think in any project that encourages outside collaborators, there's room for discussion as to how much protection is appropriate.

If we define our interfaces to be easy to use properly, people won't spend a lot of time trying to do unusual or meaningless things. To me, that's the point of design: making it easy to do the right thing and hard to do the wrong thing.

jeff-cohere:
> Should only the <struct>.c have functions to access/modify the data of the struct e.g. code change in tdympfao3D_core.c?

I think this depends on how we use the struct. We can make it easy to use TDyRegion objects easily, e.g.

```
// Loop over cells in a region and do stuff
for (int i = 0; i < region->num_cells; i++) {
  int cell_id = region->cell_ids[i];
  do_stuff(cell_id, ...);
}
```

To me, the above code is pretty easy to understand. It may be susceptible to certain pitfalls, but we may not encounter problems in practice. Personally, I think we should take a light-handed approach to encapsulation for now, and revisit each thing as we gather code examples that are particularly complicated. Then, at least, we can address specific issues instead of trying to guess how people might abuse things.

jedbrown:
The PETSc pattern is usually to use an integer range or integer array for the handles. Any "proper" object is an opaque pointer to a struct defined in a private header. There are some exceptions, such as `MatStencil`, which is basically a tuple of integers.

bishtgautam:
@jedbrown I was thinking of structs that are not visible to the application. Here is an example. In PETSc, `struct _MatStash` is an internal struct ([matimpl.h#L302](https://gitlab.com/petsc/petsc/-/blob/master/include/petsc/private/matimpl.h#L302)). The `p_Mat` struct has a member `MatStash bstash` ([matimpl.h#L452](https://gitlab.com/petsc/petsc/-/blob/master/include/petsc/private/matimpl.h#L452)). Couldn't the values of 
`mat->bstash` be modified in `matmpibaijsetvaluesblocked_()` at [mpibaij.c#L3566](https://gitlab.com/petsc/petsc/-/blob/master/src/mat/impls/baij/mpi/mpibaij.c#L3566) instead of calling `MatStashValuesRowBlocked_Private`?

jedbrown:
Yeah, that struct is just inlined; no encapsulation. The typedef could have been changed to a pointer and the struct definition moved into a `matstashimpl.h`. I suspect that wouldn't be much code churn, but also limited upside since it's already internal and not abused. `PetscLayout` is an example of a struct that is widely accessed directly (I'm not especially happy about it, but the alternative is rather verbose) and defined in a public header (legacy; it should be removable, but would be mildly disruptive to a relatively small set of users).

jeff-cohere:
I'm a big fan of the pointer-to-hidden-struct approach for encapsulation for objects with behaviors, so that's always a good option. I'm just trying to encourage us to do it where needed most (to your "limited upside" remark) because of our limited time and effort.

src/mesh/tdyregion.c
```diff
@@ -0,0 +1,38 @@
...
+  PetscFunctionReturn(0);
+}
+
+PetscBool TDyRegionAreCellsInTheSameRegion(TDyRegion *region, PetscInt cell_id_1, PetscInt cell_id_2) {
+
+  if (cell_id_1 >= region->num_cells || cell_id_2 >= region->num_cells) {
+    SETERRQ(PETSC_COMM_WORLD,PETSC_ERR_USER,"Cell ID is larger than the number of cells in the region");
+  }
+
+  PetscFunctionReturn (region->cell_ids[cell_id_1] == region->cell_ids[cell_id_2]);
```
jeff-cohere, line 37: Originally, I thought this function returned `true` if the cells with IDs `cell_id_1` and `cell_id_2` both belonged to the given region. But here, we're treating `cell_id_1` and `cell_id_2` like region indexes for cells, and returning `true` iff those two region indexes refer to the same cell. Is this what you intend?

src/mesh/tdycoremesh.c
```diff
@@ -489,6 +492,52 @@ PetscErrorCode SaveMeshGeometricAttributes(TDy tdy) {
...
+      ierr = DMCreateLocalVector(dm, &local);CHKERRQ(ierr);
+      ierr = DMGlobalToLocalBegin(dm, global, INSERT_VALUES, local); CHKERRQ(ierr);
+      ierr = DMGlobalToLocalEnd(dm, global, INSERT_VALUES, local); CHKERRQ(ierr);
+
+      // Save the region ids
+      ierr = VecGetLocalSize(local, &size_local);
+      ierr = VecGetArray(local, &p); CHKERRQ(ierr);
+
+      PetscInt ncells = size_local/num_fields;
+      PetscInt *cell_ids;
```
jeff-cohere, line 526: Since we're using C and not C++, we can actually take advantage of variable length arrays (VLAs) here. It avoids some extra allocations and is actually easier to understand.

If you want to do this, you can replace lines 526-527 with

```
PetscInt cell_ids[ncells];  // Notice that ncells is a variable, not a constant!
```

This allocates a VLA on the stack of length `ncells`. You can use it just like you'd use any other array. Then just delete line 531.

This isn't possible in C++, so folks who think that C is a proper subset of C++ aren't aware of this trick.

bishtgautam: I have made the change

jeff-cohere (approved):

PR #101, opened 2020-10-14
Fixes Neumann bc and TPF for MPFA-O

bishtgautam:
- TPF now correctly accounts for dot(n_face,dist_vector).
- For TPF, the contribution from boundary faces is not included for flux across an internal faces
- The computation of the transmissibility matrix is updated for Neumann BC.


ghammond86 (approved):

ghammond86:
Are you okay with me merging even though the patch has lower codecov coverage?

bishtgautam:
Yes, I'm okay with lower code coverage.

Issue #107, opened 2020-10-16
Only set meta data before TDySetFromOptions


bishtgautam:
- [ ] Move creation of DM
- [ ] Move allocation of memory

jeff-cohere:
I would go so far as to say "only set metadata before TDySetup".

ghammond86:
Current TDycore Order of Operations
```TDyCreate - creates and initializes the TDy struct.  No dynamic memory allocation
  TDyIOCreate
TDySetMode - sets flow mode
TDySetDiscretizationMethod - sets the discretization method
TDySet<XYZ>Function (e.g. TDySetPermeabilityFunction) - can be called after after TDy struct is created
TDySetFromOptions - harvests the metadata and overrides defaults, except for DM metadata
TDyCreateDM - can occur anywhere prior to TDySetupDiscretization
TDyDistributeDM - converts to parallel
TDySetupDiscretization - sets DM pointer, calculates grid cells, and allocates parameters arrays in TDy struct
  TDyAllocate - allocates memory for parameter arrays
Function that set values must be called or set after TDyAllocate (e.g. TDySetPermeabilityScalar, TDySetPermeabilityDiagonal, etc)
TDySetup
  TDySetupDiscretizationMethod - initializes numerical method (MPFA_O,BDM, WY, TPF)
  sets up regression linkage
```

Proposed TDycore Order of Operations
```
TDyCreate
TDySet<A>
TDySet<D>
TDySet<B>
TDySet<C>
TDySetDM
TDySetSeepageBC
TDySetFromOptions
  TDyCreateGrid
    if (!tdy->dm)
      TDyCreateDM
      DMSetUseNatural
      TDyDistributeDM
    TDyMalloc
    TDyRegressionInitialize
Set properties for each cell e.g. TDySetBlockPermeabilityValuesLocal. These call can ONLY occur after TDySetFromOptions
TDySetupNumericalMethod
  If permeability function and permeability values are not set, use default permeability value
  GMatrix, Transmissiblity matrix
TDyRunToTime
```

bishtgautam:
Fixed by #109

PR #109, opened 2020-10-19
Move TDySetFromOptions prior to DM creation; added flags to check or…

### LINKED ISSUE #106: Impose a call sequence when setting up options
After setup is called, don't allow demos to modify options. The following code should throw an error:

```
ierr = TDySetup(tdy); CHKERRQ(ierr);
ierr = TDySetDiscretizationMethod(tdy,MPFA_O); CHKERRQ(ierr);
```

ghammond86:
This pull request enforces creating and initializing members of TDycore in a specific order.  The TDycore object is now created without mallocing memory for the parameter arrays.  These arrays are allocated after TDySetFromOptions() has been called and the DM is created.  TDySetup() will not proceed unless TDySetFromOptions has been called.  A bitwise operation using tdy->setupflags enforces this order of operations.

Fixes #106  and #107

### DIFF
--- a/demo/SPE10/steady.c
+++ b/demo/SPE10/steady.c
@@ -102,7 +102,13 @@ int main(int argc, char **argv) {
   PetscErrorCode ierr;
   PetscInt dim = 2, N = 0;
   PetscReal ang = 0;
+
   ierr = TDyInit(argc, argv); CHKERRQ(ierr);
+  TDy  tdy;
+  ierr = TDyCreate(&tdy); CHKERRQ(ierr);
+  ierr = TDySetDiscretizationMethod(tdy,WY); CHKERRQ(ierr);
+  ierr = TDySetFromOptions(tdy); CHKERRQ(ierr);
+
   ierr = PetscOptionsBegin(PETSC_COMM_WORLD,NULL,"SPE Options",""); CHKERRQ(ierr);
   ierr = PetscOptionsInt ("-dim","Problem dimension","",
 			  dim,&dim,NULL); CHKERRQ(ierr);
@@ -136,14 +142,13 @@ int main(int argc, char **argv) {
   ierr = DMSetFromOptions(dm); CHKERRQ(ierr);
   ierr = DMViewFromOptions(dm, NULL, "-dm_view"); CHKERRQ(ierr);
 
-  /* Setup problem parameters */
-  TDy  tdy;
-  ierr = TDyCreateWithDM(dm,&tdy); CHKERRQ(ierr);
-  ierr = TDySetDiscretizationMethod(tdy,WY); CHKERRQ(ierr);
-  ierr = TDySetFromOptions(tdy); CHKERRQ(ierr);
-  ierr = TDySetup(tdy); CHKERRQ(ierr);
   ierr = ReadSPE10Permeability(tdy,ang); CHKERRQ(ierr);
   ierr = TDySetDirichletValueFunction(tdy,Pressure,NULL); CHKERRQ(ierr);
+
+  /* Setup problem parameters */
+  ierr = TDySetDM(dm,tdy); CHKERRQ(ierr);
+  ierr = TDyAllocate CHKERRQ(ierr);
+  ierr = TDySetup(tdy); CHKERRQ(ierr);
 
   /* Compute system */
   Mat K;
--- a/demo/SPE10/steady.c
+++ b/demo/SPE10/steady.c
@@ -107,7 +107,6 @@ int main(int argc, char **argv) {
   TDy  tdy;
   ierr = TDyCreate(&tdy); CHKERRQ(ierr);
   ierr = TDySetDiscretizationMethod(tdy,WY); CHKERRQ(ierr);
-  ierr = TDySetFromOptions(tdy); CHKERRQ(ierr);
 
   ierr = PetscOptionsBegin(PETSC_COMM_WORLD,NULL,"SPE Options",""); CHKERRQ(ierr);
   ierr = PetscOptionsInt ("-dim","Problem dimension","",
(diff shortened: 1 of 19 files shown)


demo/mpfao/mpfao.c
```diff
@@ -202,6 +202,12 @@ int main(int argc, char **argv) {
   PetscBool perturb = PETSC_FALSE;
 
   ierr = TDyInit(argc, argv); CHKERRQ(ierr);
+  TDy  tdy;
+  ierr = TDyCreate(&tdy); CHKERRQ(ierr);
```
jeff-cohere, line 206: This might be a silly question, but what do you think about merging these 4 functions, if they must be called in this order anyway? Something like (in this case):

```
ierr = TDyCreate(&tdy, RICHARDS, MPFA_O);
```
that would do the work of `TDyCreate`, `TDySetMode`, `TDySetDiscretizationMethod`, and `TDySetFromOptions` in that order, all at once? We could have utility functions for these that are hidden from view. But isn't it true that you either need all of this info in any case (or must assume some defaults if it's not provided)?

demo/mpfao/mpfao.c
```diff
@@ -222,10 +228,10 @@ int main(int argc, char **argv) {
   ierr = TDyDistributeDM(&dm); CHKERRQ(ierr);
 
   // Setup problem parameters
-  TDy  tdy;
   PetscReal gravity[3];
 
-  ierr = TDyCreateWithDM(dm,&tdy); CHKERRQ(ierr);
+  ierr = TDySetDM(dm,tdy); CHKERRQ(ierr);
+  ierr = TDyAllocate(tdy); CHKERRQ(ierr);
```
jeff-cohere, line 234: For some reason I'm having trouble finding this function. What does it do? Could it be merged into `TDySetup` below, or are you doing something in between here and that call that requires it to be here?

jeff-cohere: Ahh, I'm guessing from your description that `TDyAllocate` allocates space for parameter arrays that allow you to select functions for boundary conditions, source terms, etc. If setting the DM is the last precondition for allocating these parameter arrays, and we disallow the DM to be set for a TDy more than once, perhaps we can do this allocation at the end of `TDySetDM`.

In general, I'm trying to push for a minimum of expert knowledge in properly setting up a TDy instance. The more of these types of calls we eliminate, the less we have to explain to a basic user what's going on (and the less chance we have of messing things up).

Of course, there's a floor to the simplicity we can bestow on this process. But I'm always interested in where that floor actually is. :-)

demo/mpfao/mpfao.c
jeff-cohere: By the way, I'm thinking that both supplied arguments here (the "mode" and the discretization) can serve as defaults that can be overridden by command line options. That gives it a conceptual consistency that makes merging them together feel unforced.

jedbrown: Perhaps `TDyCreateDefault(comm, RICHARDS, MPFA_O, &tdy)` or `TDyCreateFromOptions(comm, &tdy)` and have it provide defaults and accept options.

PETSc used to have (and advocate using) a lot of convenience constructors (like `MatCreateSeqAIJ`), but they've caused more trouble than benefit. I don't know how sequential the setting of things needs to be -- I'd expect all the attributes should be possible to set in any order. Some "safer" APIs encode the "builder" process in the type system, but it's a bit cumbersome to do in C.

In any case, functions with lots of arguments are harder to understand (hard to remember what goes where) and more brittle in that any new argument breaks all existing use. In contrast, a sequence of `TDySetX` calls are easy to read and don't break code so long as defaults are available.

ghammond86: We can merge, but we will still need those two separate functions on the library side.  Eventually, there will only one driver with those function calls.

demo/mpfao/mpfao.c
ghammond86: I agree.  I have renamed TDySetDM -> TDySetupDiscretization and moved the portion of TDyAllocate that calculates the geometry within.  I then embed the call to TDyAllocate at the end of TDySetupDiscretization.  The calls to TDyAllocate have been removed from the demos.

demo/SPE10/steady.c
```diff
@@ -136,15 +142,13 @@ int main(int argc, char **argv) {
...
-  TDy  tdy;
-  ierr = TDyCreateWithDM(dm,&tdy); CHKERRQ(ierr);
-  ierr = TDySetDiscretizationMethod(tdy,WY); CHKERRQ(ierr);
-  ierr = TDySetFromOptions(tdy); CHKERRQ(ierr);
-  ierr = TDySetup(tdy); CHKERRQ(ierr);
   ierr = ReadSPE10Permeability(tdy,ang); CHKERRQ(ierr);
   ierr = TDySetDirichletValueFunction(tdy,Pressure,NULL); CHKERRQ(ierr);
 
+  /* Setup problem parameters */
+  ierr = TDySetupDiscretization(dm,tdy); CHKERRQ(ierr);
```
bishtgautam, line 149: Shouldn't this be `TDySetDM`?

bishtgautam: @ghammond86 I just saw [your comment](https://github.com/TDycores-Project/TDycore/pull/109#discussion_r508645420) about `TDySetDM -> TDySetupDiscretization`

bishtgautam:
@ghammond86, I had the following workflow in mind: 

```
TDyCreate (&tdy);

// Set any options you want
TDySetA(tdy, ...);
TDySetB(tdy, ...);
TDySetC(tdy, ...);
TDySetD(tdy, ...);

// Finally
TDySetFromOptions();  // This function will allocate memory, 
                      // compute transmissibility matrix for MPFA-O, etc

```

If one of the options called prior to `TDySetFromOptions()` was `TDySetDM`, TDycore will use the user-provided DM, otherwise, it will create the DM from a file or a box mesh.

jeff-cohere:
> @ghammond86, I had the following workflow in mind:
> 
> ```
> TDyCreate (&tdy);
> 
> // Set any options you want
> TDySetA(tdy, ...);
> TDySetB(tdy, ...);
> TDySetC(tdy, ...);
> TDySetD(tdy, ...);
> 
> // Finally
> TDySetFromOptions();  // This function will allocate memory, 
>                       // compute transmissibility matrix for MPFA-O, etc
> ```
> 
> If one of the options called prior to `TDySetFromOptions()` was `TDySetDM`, TDycore will use the user-provided DM, otherwise, it will create the DM from a file or a box mesh.

If we can eliminate the ordering requirements for A, B, C, D above, it sounds great to me.

bishtgautam:
I forgot to mention in my previous comment that there would be no prescribed order of options.

bishtgautam (approved):

PR #114, opened 2020-11-04
Fixes namespace

### LINKED ISSUE #113: Use internal (or static) linkage for anything that should not be in the public API
Based on @jedbrown's [comment in #105](https://github.com/TDycores-Project/TDycore/issues/105#issuecomment-710475904).

bishtgautam:
Multiple internal functions are changed from `PETSC_EXTERN` to `PETSC_INTERN`

Fixes #113


demo/richards/richards_driver.c
```diff
@@ -37,8 +35,10 @@ int main(int argc, char **argv) {
   ierr = TDySetFromOptions(tdy); CHKERRQ(ierr);
 
   ierr = TDyDriverInitializeTDy(tdy); CHKERRQ(ierr);
-  if (!rank) tdy->io->io_process = PETSC_TRUE;
-  tdy->io->print_intermediate = print_intermediate;
+  if (!rank) {
```
jeff-cohere, line 38: Not a big deal, but if you are using `!rank` here to mean `rank 0`, then `if (rank == 0)` is more clear.

include/private/tdympfaoimpl.h
```diff
@@ -4,7 +4,6 @@
 #include <petsc.h>
 
 PETSC_EXTERN PetscErrorCode TDyMPFAOInitialize(TDy);
```
jeff-cohere, line 6: Just to check: do you want external linkage on this function? Does one of the drivers use it?

include/tdycore.h
```diff
@@ -77,19 +77,18 @@ PETSC_EXTERN PetscErrorCode TDyInit(int argc, char* argv[]);
...
 PETSC_EXTERN PetscErrorCode TDyFinalize(void);
 
-PETSC_EXTERN PetscErrorCode TDyCreate(TDy *tdy);
-PETSC_EXTERN PetscErrorCode TDySetMode(TDy tdy, TDyMode mode);
-PETSC_EXTERN PetscErrorCode TDySetDiscretizationMethod(TDy tdy,
-                                                       TDyMethod method);
-PETSC_EXTERN PetscErrorCode TDySetDM(TDy tdy,DM dm);
-PETSC_EXTERN PetscErrorCode TDySetFromOptions(TDy tdy);
-PETSC_EXTERN PetscErrorCode TDySetupNumericalMethods(TDy tdy);
+PETSC_EXTERN PetscErrorCode TDyCreate(TDy*);
```
jeff-cohere, line 80: I'm not sure we want to get rid of these parameter names for functions that belong to TDycore's public API. If we adopt a Doxygen-style approach to annotating source code, we would describe these parameter names in the annotations. Of course, we can annotate the implementations instead of the declarations, but many folks are used to browsing header files to see what a library can do.

jeff-cohere (approved):
Looks good. I left some comments that we can discuss. The only thing I might change is the omission of all the parameter names from function declarations in headers, because they're arguably a part of our reference documentation approach.

demo/richards/richards_driver.c
jedbrown: I don't disagree, but this is a PETSc convention that Barry felt very strongly about so we (PETSc developers) always write `!rank`.

include/private/tdympfaoimpl.h
bishtgautam: Good catch. I have fixed this now.

demo/richards/richards_driver.c
bishtgautam: I simply followed the existing code and don't have a strong opinion either way. One more reason to have a style guide.

include/tdycore.h
bishtgautam: I'm here following PETSc's developer manual. Another reason for us to have our own style guide.

jeff-cohere: Sounds like we need a style guide!

jedbrown: The rationale for the PETSc style guide is that the compiler doesn't check the names appearing here so they can become outdated and thus misleading.

jeff-cohere: That makes sense for an approach that doesn't annotate header files. If we decide to annotate our functions in their implementations instead of headers, then this approach is fine.

PR #117, opened 2020-11-06
Refactors code for material properties, conditions, and constitutive models

bishtgautam:
The refactored code consolidates code in the following files:
1. `tdymaterialproperties.c`: For soil properties such as permeability, porosity, thermal conductivity, etc
2. `tdycondition.c`: For setting Dirichlet and source-sink condition
3. `tdycharacteristiccurves.c`: For relative permeability and saturation functions


bishtgautam:
Few outstanding issues that can be addressed in here or a future PR:
1. The `tdycharacteristiccurves.c` and `tdyeos.c` could be moved into a new `src/constitutive_model`.
2. The two supported user-defined function types (with and without typedef) could be consolidated into one. (cc: @jedbrown)

jeff-cohere:
Some of the names of these functions/files are pretty long. How do we feel about introducing some abbreviations, or omitting some words? For example, in physics I'm used to discussing a "heat capacity" (had to check to make sure "specific heat capacity" was the same). And `MatProps` or `MaterialProps` may convey the same thing as `MaterialProperties`. Same for `CharCurves`, if that's how we want to refer to relative permeability and saturation functions.

bishtgautam:
Abbreviations would certainly be good. Let's create a list of long names. I will try to work on creating such a list later today.

include/private/tdycoreimpl.h
```diff
@@ -61,26 +61,27 @@ struct _p_TDy {
...
-  PetscReal *matprop_m, *matprop_n, *matprop_alpha;
+  PetscReal *cc_sr;              /* residual saturation (min) [1] */
+  PetscReal *cc_m;               /* parameter used in saturation and relative permeability function [-]*/
+  PetscReal *cc_n;               /* parameter used in Gardner saturation function [-] */
+  PetscReal *cc_alpha;           /* parameter used in VanGenuchten saturation function [-] */
+  PetscReal *S,                  /* saturation for each cell */
+            *dS_dP,              /* derivative of saturation wrt pressure for each cell [Pa^-1] */
+            *d2S_dP2,            /* second derivative of saturation wrt pressure for each cell [Pa^-2] */
+            *dS_dT;              /* derivate of saturation wrt to temperature for each cell [K^-1] */
+  PetscReal *Kr, *dKr_dS;        /* relative permeability for each cell [1] */
```
bishtgautam, line 84: Create a struct for this say `struct _p_TDyCharCurves`. This will avoid adding a prefix `cc_`.

include/private/tdycoreimpl.h
```diff
@@ -61,26 +61,27 @@ struct _p_TDy {
...
-  PetscReal *Cr;                 /* specific heat capacity for rock [J/(kg-K)] */
-  PetscReal *rhor;               /* rock density [kg/m3] */
-
+  PetscReal *matprop_K,
+            *matprop_K0;         /* permeability tensor (cell,intrinsic) for each cell [m2] */
+  PetscReal *matprop_porosity;   /* porosity for each cell [1] */
+  PetscReal *matprop_Kappa,
+            *matprop_Kappa0;     /* thermal conductivity tensor (cell,intrinsic) for each cell [W/(K-m)] */
+  PetscReal *matprop_Cr;         /* specific heat capacity for rock [J/(kg-K)] */
+  PetscReal *matprop_rhor;       /* rock density [kg/m3] */
```
bishtgautam, line 71: Create a struct for this, say `struct _p_TDyMatProps`. This will avoid adding a prefix `matprop_`.

include/private/tdycharacteristiccurvesimpl.h
```diff
@@ -1,13 +1,21 @@
...
+  REL_PERM_FUNC_MUALEM=1
+} TDyRelPermFuncType;
+
 typedef enum {
   SAT_FUNC_GARDNER=0,
   SAT_FUNC_VAN_GENUCHTEN=1
 } TDySatFuncType;
 
+PETSC_INTERN void RelativePermeability_Mualem(PetscReal,PetscReal,PetscReal*,PetscReal*);
+PETSC_INTERN void RelativePermeability_Irmay(PetscReal,PetscReal,PetscReal*,PetscReal*);
```
bishtgautam, line 17: Change this to `RelPermMualem` and ``RelPermIrmay`

bishtgautam: On second thought, let's leave it as is and the name will be changed after we have developed a style guide.

bishtgautam:
@jeff-cohere, Apart from creating new structs, we should go with your suggestion of `MaterialProps` and `CharCurves`  (`MatProp` might be confused with PETSc's `Mat`). I will make those changes if that sounds good to you.

include/private/tdycoreimpl.h
jedbrown: I agree even if it's an anonymous struct.
```c
struct Outer {
  PetscReal lots_of_fields;
  struct {      // Characteristic curves
    PetscReal *sr;
    ...
  } cc;
  PetscReal other_fields;
};
```
Which would be accessed as `ctx->cc.sr`.

FWIW, PETSc uses that `_p_StructName` for `PetscObject`s (which carry a communicator, reference counting and destruction semantics, etc.).

jedbrown:
It isn't really my place, but abbreviated names have a learning curve and sustained cognitive load so I'd stick with non-abbreviated names unless the characters become really expensive (like users need to type them repeatedly or it makes lines too long in parts of the code with nontrivial logic).

jeff-cohere:
> It isn't really my place, but abbreviated names have a learning curve and sustained cognitive load so I'd stick with non-abbreviated names unless the characters become really expensive (like users need to type them repeatedly or it makes lines too long in parts of the code with nontrivial logic).

Good point. Still, there are plenty of abbreviations in most libraries already, and not all of them are equally opaque. And there are some "empty calories": I don't see anyone being tripped up by `MaterialProps` instead of `MaterialProperties`. And if `SpecificHeatCapacity` actually means `SpecificHeat`, the latter is both shorter and less confusing to people who have taken the vanilla version of thermodynamics.

We should certainly select them with caution (`MatProps` is not great, as Gautam has pointed out), but abbreviations are just another tool with another set of tradeoffs. And we're going to need documentation for function arguments anyway.

bishtgautam:
> learning curve and sustained cognitive load

@jedbrown's point is good.  Let's make the following change:
- `MaterialProperties` -> `MaterialProp`
- `SpecificHeatCapacity` -> `SpecificHeat`

but leave `CharacteristicCurves` unchanged because `CharCurves` feels like inside baseball.

include/private/tdycoreimpl.h
bishtgautam: Btw, what does `_p` refer to? PETSc or private or something else?

jedbrown: "private" or "PETSc Object" (`_n_` has been used for typedefs that are not PetscObjects). It's actually bad style because identifiers starting with `_` are reserved. We append `_Private` in libCEED.

bishtgautam:
@jeff-cohere This is ready for your review.

include/private/tdycoreimpl.h
```diff
@@ -61,33 +63,20 @@ struct _p_TDy {
...
-  PetscReal *Cr;                 /* specific heat capacity for rock [J/(kg-K)] */
-  PetscReal *rhor;               /* rock density [kg/m3] */
-
-  PetscInt *SatFuncType;         /* type of saturation function */
-  PetscInt *RelPermFuncType;     /* type of relative permeability */
-
-  PetscReal *matprop_m, *matprop_n, *matprop_alpha;
+  MaterialProp matprop;
+
+  /* characteristic curver parameters */
```
jeff-cohere, line 69: curve, not curver

src/mpfao/2D/tdympfao2D.c
```diff
@@ -69,7 +70,7 @@ PetscErrorCode TDyComputeGMatrixFor2DMesh(TDy tdy) {
     // extract permeability tensor
     for (ii=0; ii<dim; ii++) {
       for (jj=0; jj<dim; jj++) {
-        K[ii][jj] = tdy->K[icell*dim*dim + ii*dim + jj];
+        K[ii][jj] = matprop->K[icell*dim*dim + ii*dim + jj];
```
jeff-cohere, line 73: If we intend the fields of a MaterialProp structure to be directly manipulated, I don't think we should make its underlying type a pointer. The pointer pattern makes sense when you manipulate the data of an "object" only through functions that belong to its interface.

The reason is this: here we see that a variable is declared as what seems to be a scalar type, and then it's treated as a pointer (with `->` to access its fields). This looks awfully strange to a C programmer. If it's really just a collection of fields to be accessed freely without an interface, I recommend that we use what we're calling `_MaterialProp` as `MaterialProp`.

src/mpfao/3D/tdympfao3D_th_ts.c
```diff
@@ -91,22 +92,22 @@ PetscErrorCode TDyMPFAOIFunction_Vertices_3DMesh_TH(Vec Ul, Vec R, void *ctx) {
       if (TtimesP[irow] < 0.0) { // up ---> dn
          // Is the cell_id_up an internal or boundary cell?
         if (cell_id_up >= 0) {
-          ukvr = tdy->Kr[cell_id_up]/tdy->vis[cell_id_up];
+          ukvr = cc->Kr[cell_id_up]/tdy->vis[cell_id_up];
```
jeff-cohere, line 95: My comment for `MaterialProp` also applies here.

src/mpfao/2D/tdympfao2D.c
jeff-cohere: If we redefine `MaterialProp` as just a struct, line 64 changes to

```
MaterialProp* matprop = tdy->matprop;
```
This makes it clear that we're using a pointer, and that the fields should be accessed via `->`.

include/private/tdymaterialpropertiesimpl.h
```diff
@@ -0,0 +1,26 @@
+#if !defined(TDYMATERIALPROPERTIES_H)
+#define TDYMATERIALPROPERTIES_H
+
+#include <petsc.h>
+
+typedef struct _MaterialProp MaterialProp;
```
jeff-cohere, line 6: Instead of having a separate typedef, you could just include it in the struct's definition:

```
typedef struct {
    PetscReal *K, *K0;
    PetscReal *porosity;
    PetscReal *Kappa, *Kappa0;
    PetscReal *Cr;
    PetscReal *rhosoil;
} MaterialProp;
```

This results in only having one symbol related to `MaterialProp`, which is a basic struct not a "class"-style struct with a private definition. Not a big deal, but simpler.

jeff-cohere: You can do the same thing for `CharacteristicCurve` as well. But like I said, it's not a big deal, so I'm fine with them as is.

src/tdycharacteristiccurves.c
```diff
@@ -0,0 +1,283 @@
...
+  if (cc->dS_dT          ) { ierr = PetscFree(cc->dS_dT          ); CHKERRQ(ierr); }
+  if (cc->sr             ) { ierr = PetscFree(cc->sr             ); CHKERRQ(ierr); }
+  if (cc->m              ) { ierr = PetscFree(cc->m              ); CHKERRQ(ierr); }
+  if (cc->n              ) { ierr = PetscFree(cc->n              ); CHKERRQ(ierr); }
+  if (cc->alpha          ) { ierr = PetscFree(cc->alpha          ); CHKERRQ(ierr); }
+  
+  PetscFunctionReturn(0);
+}
+
+PetscErrorCode TDySetResidualSaturationValuesLocal(TDy tdy, PetscInt ni, const PetscInt ix[], const PetscScalar y[]){
```
jeff-cohere, line 50: FYI: In these kinds of functions, you can actually place `ni` within the square brackets for the `ix` and `y` arguments, to tell the compiler to expect arrays of this size.

src/tdycore.c
```diff
@@ -212,67 +211,57 @@ PetscErrorCode TDyMalloc(TDy tdy) {
...
   nc   = cEnd-cStart;
-  ierr = PetscMalloc(dim*dim*nc*sizeof(PetscReal),&(tdy->K0)); CHKERRQ(ierr);
-  ierr = PetscMalloc(dim*dim*nc*sizeof(PetscReal),&(tdy->K )); CHKERRQ(ierr);
-  ierr = PetscMalloc(nc*sizeof(PetscReal),&(tdy->porosity)); CHKERRQ(ierr);
-  ierr = PetscMalloc(nc*sizeof(PetscReal),&(tdy->Kr)); CHKERRQ(ierr);
-  ierr = PetscMalloc(nc*sizeof(PetscReal),&(tdy->dKr_dS)); CHKERRQ(ierr);
-  ierr = PetscMalloc(nc*sizeof(PetscReal),&(tdy->S)); CHKERRQ(ierr);
-  ierr = PetscMalloc(nc*sizeof(PetscReal),&(tdy->dS_dP)); CHKERRQ(ierr);
-  ierr = PetscMalloc(nc*sizeof(PetscReal),&(tdy->d2S_dP2)); CHKERRQ(ierr);
+  //ierr = PetscMalloc(dim*dim*nc*sizeof(PetscReal),&(tdy->matprop_K0)); CHKERRQ(ierr);
```
jeff-cohere, line 214: Any reason for just commenting these out? Do you expect to reintroduce them at some point?

bishtgautam: This is an oversight. Will fix it.

include/private/tdymaterialpropertiesimpl.h
bishtgautam: I will make suggested changes.

jeff-cohere (approved):
Looks good. Much better organization. I've left some comments to think about, but nothing gets in the way of approval.

bishtgautam:
This is ready to go.

PR #122, opened 2020-11-12
Changed invocation of MatSetValuesLocal in 3D MPFAO TS.

jeff-cohere:
Now we set all row values simultaneously. Regression tests pass.
I haven't profiled the performance to compare, but this should be
a bit faster.

I also moved variable declarations in the relevant function within
the scope of their usage. This is a modern way of declaring variables
that aids learning and debugging, because it makes clear how variables
are used instead of piling them at the top of a function. Unfortunately,
Fortran doesn't support it, so it might look a bit weird for some.

### DIFF
--- a/src/mpfao/3D/tdympfao3D_ts.c
+++ b/src/mpfao/3D/tdympfao3D_ts.c
@@ -195,55 +195,39 @@ PetscErrorCode TDyMPFAOIFunction_3DMesh(TS ts,PetscReal t,Vec U,Vec U_t,Vec R,vo
 PetscErrorCode TDyMPFAOIJacobian_Vertices_3DMesh(Vec Ul, Mat A, void *ctx) {
 
   TDy tdy = (TDy)ctx;
-  TDy_mesh *mesh;
-  TDy_cell *cells;
-  TDy_face *faces;
-  TDy_vertex *vertices;
-  DM dm;
-  PetscInt ivertex, vertex_id;
-  PetscInt npitf_bc, nflux_in;
-  PetscInt cell_id, cell_id_up, cell_id_dn;
-  PetscInt irow, icol;
-  PetscInt dim;
-  PetscReal gz;
-  PetscReal *p;
-  PetscReal ukvr, den;
-  PetscReal dukvr_dPup, dukvr_dPdn, Jac;
-  PetscReal dden_dPup, dden_dPdn;
-  PetscReal T;
-  PetscScalar *TtimesP_vec_ptr;
-
-  PetscErrorCode ierr;
 
   PetscFunctionBegin;
   TDY_START_FUNCTION_TIMER()
 
-  mesh     = tdy->mesh;
-  cells    = &mesh->cells;
-  faces    = &mesh->faces;
-  vertices = &mesh->vertices;
-  dm       = tdy->dm;
+  TDy_mesh *mesh  = tdy->mesh;
+  TDy_cell *cells = &mesh->cells;
+  TDy_face *faces = &mesh->faces;
+  TDy_vertex *vertices = &mesh->vertices;
+  DM dm = tdy->dm;
 
-  ierr = DMGetDimension(dm,&dim); CHKERRQ(ierr);
+  PetscInt dim;
+  PetscErrorCode ierr = DMGetDimension(dm,&dim); CHKERRQ(ierr);
 
+  PetscReal* p;
   ierr = VecGetArray(Ul,&p); CHKERRQ(ierr);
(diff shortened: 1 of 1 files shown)


src/mpfao/3D/tdympfao3D_ts.c
```diff
@@ -271,27 +255,28 @@ PetscErrorCode TDyMPFAOIJacobian_Vertices_3DMesh(Vec Ul, Mat A, void *ctx) {
     // For k not equal to i and j, jacobian is given as:
     // d(fluxm_ij)/dP_k =   rho_ij       + (kr/mu)_{ij,upwind}         *   T_ik  *  (1+d(rho_k)/dP_k*g*z
     //
-    for (irow=0; irow<nflux_in + npitf_bc; irow++) {
+    for (PetscInt irow=0; irow<nflux_in + npitf_bc; irow++) {
```
bishtgautam, line 258: I recall getting an error with this on Linux, but I need to double-check to confirm it.

jeff-cohere: This can give you a warning (or an error, depending on your compiler flags) if you use it for a variable with the same name as one in the outer scope. But otherwise it's fine and in fact recommended for loop variables that aren't needed outside their loops.

jedbrown: It's a c99 feature, but even PETSc allows this (since earlier this year). I prefer declarations at innermost scope (makes it easier to read and refactor code).

bishtgautam: > if you use it for a variable with the same name as one in the outer scope.

Yes, that was the case

bishtgautam (approved):

src/mpfao/3D/tdympfao3D_ts.c
```diff
@@ -324,19 +309,26 @@ PetscErrorCode TDyMPFAOIJacobian_Vertices_3DMesh(Vec Ul, Mat A, void *ctx) {
...
+        PetscInt cell_id = vertices->internal_cell_ids[vOffsetCell + icol];
 
-        T = tdy->Trans[vertex_id][irow][icol];
+        PetscReal T = tdy->Trans[vertex_id][irow][icol];
 
+        PetscReal gz;
         ierr = ComputeGtimesZ(tdy->gravity,cells->centroid[cell_id].X,dim,&gz); CHKERRQ(ierr);
 
+        PetscReal Jac;
         if (cell_id_up>-1 && cell_id == cell_id_up) {
```
bishtgautam, line 332: A random thought. If I understand vectorization, I think this code can also be vectorized as the conditionals are based on cell-ids and they will not change during time-stepping. For a given `icol`, the `J` can be written as:
```
J =  aJ^1_up + bJ^2_up + cJ^3_up + dJ^1_dn + eJ^2_dn + fJ^3_dn + J^4
```
where 
- `J^1_up = dden_dPup * ukvr       * TtimesP[irow]` (does not change for `icol`)
- `J^2_up = den       * dukvr_dPup * TtimesP[irow]` (does not change for `icol`)
- `J^3_up = den       * ukvr       * T * dden_dPup*gz ` (does change for `icol`)
- `J^1_dn = dden_dPdn * ukvr       * TtimesP[irow]` (does not change for `icol`)
- `J^2_dn = den       * dukvr_dPdn * TtimesP[irow]` (does not change for `icol`)
- `J^3_dn = den       * ukvr       * T * dden_dPdn*gz ` (does change for `icol`)
- `J^3.  = den       * ukvr       * T` (does change for `icol`)

The coefficients `a`-`f` for various Jacobian terms will be saved during initialization and will be given by:
```
J_up    = 1 x J^1_up + 1 x J^2_up + 1 x J^3_up + 0 x J^1_dn + 0 x J^2_dn + 0 x J^3_dn + 1 x J^4
J_dn    = 0 x J^1_up + 0 x J^2_up + 0 x J^3_up + 1 x J^1_dn + 1 x J^2_dn + 1 x J^3_dn + 1 x J^4
J_other = 0 x J^1_up + 0 x J^2_up + 0 x J^3_up + 0 x J^1_dn + 0 x J^2_dn + 0 x J^3_dn + 1 x J^4
```

bishtgautam (approved):

src/mpfao/3D/tdympfao3D_ts.c
jeff-cohere: I've recorded this link in a vectorization-related issue. For now, I'll get this PR merged.

PR #131, opened 2020-11-19
Added some Doxygen markup examples to the style guide and to the TDyRegion type.

### LINKED ISSUE #128: Add example of comment for doxygen

jeff-cohere:
To get us started documenting our API, I added what I recommend we use in Doxygen's markup.

On an almost-unrelated issue: I confess I'm still a bit confused about how the `TDyRegion` type is used in practice, and the function that adds cells to it could be improved. But it's probably not a high priority. @bishtgautam , let me know if you are bored and want to discuss it.

Closes #128

### DIFF
--- a/doc/sphinx/source/styleguide.rst
+++ b/doc/sphinx/source/styleguide.rst
@@ -140,7 +140,7 @@ The destructor function is named after the class with a ``Destroy`` suffix,
 accepts a pointer to the instance of the class to be destroyed, and returns an
 error code indicating whether an error was encountered. For example:::
 
-    PetscErrorCode WashingMachineDestroy(WashingMachine* wm);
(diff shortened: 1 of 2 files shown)


bishtgautam (approved):

bishtgautam:
@jeff-cohere This fell off my radar. With the Github actions in #136, will be able also able to generate the Doxygen pages automatically and host them on github?

jeff-cohere:
I looked into this a bit, and I think how it works is that we refer to Doxygen elements within the Restructured Text files, and they get inserted and beautified. So it's not as automatic as maybe we'd like, but it looks nice and is easy to use. We can also host the Doxygen HTML files separately if we want.

Issue #135, opened 2020-11-25
Ensure that ghost cells exist for nodes in MPFA-O


bishtgautam:
@jeff-cohere I think one solution could be to save information about "additional" cells that are shared by a vertex but are not within local+ghost cells that a graph partitioner will provide information about. To do so, the following two functions need to be updated:
- [SaveMeshGeometricAttributes](https://github.com/TDycores-Project/TDycore/blob/master/src/mesh/tdycoremesh.c#L380)
- [SaveMeshConnectivityInfo](https://github.com/TDycores-Project/TDycore/blob/master/src/mesh/tdycoremesh.c#L541)

In order to save information about additional cells, memory needs to be correctly allocated in [TDyAllocateMemoryForMesh](https://github.com/TDycores-Project/TDycore/blob/master/src/mesh/tdycoremesh.c#L300). Memory allocation for the mesh and saving attributes about the mesh are implemented at [tdympfao.c#L310](https://github.com/TDycores-Project/TDycore/blob/master/src/mpfao/tdympfao.c#L310) and [tdympfao.c#L407](https://github.com/TDycores-Project/TDycore/blob/master/src/mpfao/tdympfao.c#L407), respectively.

What do you think?

jeff-cohere:
I think as long as these functions are called by every process at the same time, this is a relatively simple approach that could work very well. I'll take a look.

jeff-cohere:
From the mailing list:

-----

If c0 is owned by MPI rank 0 and all other cells are owned by different MPI ranks, is there a stencil type that will lead to c1-c5 being ghost cells on rank 0? Essentially, I need information about all cells sharing a vertex on rank 0.

A Plex is a parallel collection of local meshes. Each local mesh is guaranteed to contain the closure (transitive closure of cone) of each point. However, the local mesh is _not_ guaranteed to contain the supports of points.

Thus, if you need cell neighbors you will need to DMPlexDistribute() with an overlap argument of 1. Then all those cells will be in the local mesh. You can distinguish an "owned" cell from a "ghost" cell by searching for it in the pointSF (DMGetPointSF). I do this using PetscSFGetGraph() and searching in the leaves, which are sorted, so you can use PetscFindInt().

jeff-cohere:
Hey @bishtgautam , do we have a test in mind that relies on nodal control volumes being correct for a parallel run? Is there an existing serial MPFA-O test that doesn't yet work in parallel that would benefit from having the nodal ghost cells available?

[...]

PR #137, opened 2020-12-08
Cleaning up tdycoremesh.c to limit "surface area" of variables.

jeff-cohere:
This is just a source cleanup of tdycoremesh.c. I removed a lot of the Fortran-like "walls of variables" at the beginnings of functions, moving them to where these variables are used. As explained in our style guide, this reduces the number of places that they can (a) contain irrelevant data and (b) cause problems.

Pertains to #135

### DIFF
--- a/src/mesh/tdycoremesh.c
+++ b/src/mesh/tdycoremesh.c
@@ -284,9 +284,10 @@ PetscErrorCode TDyAllocateMemoryForMesh(TDy tdy) {
 
   DM dm = tdy->dm;
   TDyMesh *mesh = tdy->mesh;
+  PetscErrorCode ierr;
 
   PetscInt dim;
-  PetscErrorCode ierr = DMGetDimension(dm, &dim); CHKERRQ(ierr);
+  ierr = DMGetDimension(dm, &dim); CHKERRQ(ierr);
 
   if (dim!= 2 && dim!=3 ) {
     SETERRQ(PETSC_COMM_WORLD,PETSC_ERR_SUP,"Only 2D and 3D grids are supported");
@@ -361,10 +362,11 @@ PetscErrorCode SaveMeshGeometricAttributes(TDy tdy) {
   TDyVertex     *vertices = &mesh->vertices;
   TDyEdge       *edges = &mesh->edges;
   TDyFace       *faces = &mesh->faces;
+  PetscErrorCode ierr;
 
   // Determine the number of cells, edges, and vertices of the mesh
   PetscInt c_start, c_end;
-  PetscErrorCode ierr = DMPlexGetHeightStratum(dm, 0, &c_start, &c_end); CHKERRQ(ierr);
+  ierr = DMPlexGetHeightStratum(dm, 0, &c_start, &c_end); CHKERRQ(ierr);
 
   PetscInt e_start, e_end;
   ierr = DMPlexGetDepthStratum(dm, 1, &e_start, &e_end); CHKERRQ(ierr);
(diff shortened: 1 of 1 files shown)


jeff-cohere:
By the way, it looks like there are a lot of mesh-related functions. Let me know if you think we would benefit from making the mesh into a more formal "class" type.

src/mesh/tdycoremesh.c
```diff
@@ -387,110 +361,120 @@ PetscErrorCode SaveMeshGeometricAttributes(TDy tdy) {
...
-  PetscInt       eStart, eEnd;
-  PetscInt       fStart, fEnd;
-  PetscInt       pStart, pEnd;
-  PetscInt       icell, iedge, ivertex, ielement, iface, d;
-  PetscErrorCode ierr;
 
-  ierr = DMGetDimension(dm, &dim); CHKERRQ(ierr);
+  // Determine the number of cells, edges, and vertices of the mesh
+  PetscInt c_start, c_end;
+  PetscErrorCode ierr = DMPlexGetHeightStratum(dm, 0, &c_start, &c_end); CHKERRQ(ierr);
```
bishtgautam, line 367: Because `ierr` is used in subsequent lines, how about instead using the following?
```
PetscErrorCode ierr;
ierr = DMPlexGetHeightStratum(dm, 0, &c_start, &c_end); CHKERRQ(ierr); 
```

src/mesh/tdycoremesh.c
```diff
@@ -548,66 +532,59 @@ PetscErrorCode SaveMeshConnectivityInfo(TDy tdy) {
...
-  PetscReal      v_1[3], v_2[3];
-  PetscInt       d;
-  PetscErrorCode ierr;
 
+  PetscInt nverts_per_cell = tdy->ncv;
 
-  ierr = DMGetDimension(dm, &dim); CHKERRQ(ierr);
+  // Determine the number of cells, edges, and vertices of the mesh
+  PetscInt c_start, c_end;
+  PetscErrorCode ierr = DMPlexGetHeightStratum(dm, 0, &c_start, &c_end); CHKERRQ(ierr);
```
bishtgautam, line 540: Same as the previous comment about `ierr`

src/mesh/tdycoremesh.c
```diff
@@ -781,77 +771,67 @@ PetscErrorCode UpdateCellOrientationAroundAVertex(TDy tdy, PetscInt ivertex) {
...
 
     ids[count]              = icell;
     idx[count]              = count;
     subcell_ids[count]      = vertices->subcell_ids[vOffsetSubcell + i];
     is_cell[count]          = PETSC_TRUE;
     is_internal_edge[count] = PETSC_FALSE;
 
-    ierr = ComputeTheta(x, y, &theta[count]);
-
+    PetscErrorCode ierr = ComputeTheta(x, y, &theta[count]); CHKERRQ(ierr);
```
bishtgautam, line 800: Same as the previous comment about `ierr`

bishtgautam (commented):
The changes look good to me except the changes to `ierr` that I have commented in my review. There are a few more places where the same comment regarding `ierr` is applicable, but I haven't left a message in all locations.

bishtgautam:
>making the mesh into a more formal "class" type.

Can you create a pseudo code for the change you have in mind?

jeff-cohere:
Okay, I fixed the `ierr` declarations.

I'll poke around and see whether I believe the mesh needs an interface. I was just noticing the large number of functions managing various parts of the mesh's state. If it seems to me that the mesh deserves its own interface, I'll put together some pseudo-code. But that can wait for another PR. This set of changes is large enough that I'd rather get them in quick.

bishtgautam (approved):

PR #139, opened 2020-12-22
Introduced a CMocka unit testing proof of concept

jeff-cohere:
This little guy allows us to use a pre-installed CMocka library whose location is defined by the CMOCKA_INCLUDE_DIR and CMOCKA_LIBRARY_DIR environment variables. See `gmakefile` for details. If either of these environment variables is not defined, unit tests are deactivated.

Briefly, you can just type

```
make unit-test
```

to build and run the unit tests only. They're also included in the `test` target.

As you can see in the unit test (`src/tests/test_tdyinit.c`), I've placed some simple checks that MPI does in fact work within CMocka. To run the unit tests with a specific number of processes, you can use (e.g.)

```
env MPI_NPROC=4 make unit-test
```

This is just a starting point. I'd love to hear thoughts on how we make this easier to use.

Closes #138

PR #141, opened 2020-12-28
Created a GitHub Action to replace Travis CI.

jeff-cohere:
Travis CI is adopting a less generous pricing model, and GitHub Actions seem to be doing a better job at CI these days. In fact, we're already using a GitHub Action to publish our docs. On Jed's recommendation (see #139), I thought I'd play around with a GitHub Action that replaces Travis CI. This might be a little noisy till I get everything right, so apologies to those who closely monitor the repo.

This PR adds such a GitHub action, and removes the existing Travis CI logic.


.github/workflows/auto_pr_test.yml
```diff
@@ -0,0 +1,103 @@
...
+        --download-hdf5 \
+        --download-metis \
+        --download-parmetis \
+        --download-exodusii \
+        --download-netcdf \
+        --download-pnetcdf \
+        --download-zlib \
+        --download-fblaslapack \
+        --download-mpich=http://www.mpich.org/static/downloads/3.2/mpich-3.2.tar.gz
+        make all
```
jedbrown, line 69: This will take a while. You could use an image that already has everything (libCEED uses `jedbrown/petsc:latest`, which is updated to track PETSc master periodically and there have been requests to automate it from PETSc upstream). Alternatively, use apt to install `libblis-dev` and `liblapack-dev`.

Also, why `--with-shared-libraries=0`? That increases link time because we have several executables.

In PETSc CI, we often build with debug, but with `COPTFLAGS=-O` (modest optimization; no debugging) because we aren't actually attaching a debugger to jobs in CI and thus no need to pay for debug symbols.

jedbrown (approved):
Looks great. Some comments on speeding it up.

.github/workflows/auto_pr_test.yml
```diff
@@ -0,0 +1,103 @@
...
+        make test-mpfao
+        make test-steady
+        make test-steadyf90
+        make test-transient
+        make test-transient-mpfao
+        make test-transient-mpfaof90
+        make test-transient-snes-mpfaof90
+        make test-transient-th-mpfao
+        make test-richards
+        make test-th
```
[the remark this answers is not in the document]
jeff-cohere: I agree. I lifted this directly from the Travis workflow, and wasn't sure whether it was covered by our current `test` target. I'll see if anything's missing there.

.github/workflows/auto_pr_test.yml
jeff-cohere: Thanks for the image! That will help quite a bit.

I usually default to static linking because it's the only thing that Always Works (tm). In this case, though, we were doing this in Travis. Anyway, I agree about the link times and am happy to change it to dynamic.

I'll also add the `COPTFLAGS=0` argument.

jedbrown: I built image this way so if you use it, you don't need to configure anything with PETSc.
https://github.com/jedbrown/jedbrown-dockerfiles/blob/main/petsc/Dockerfile

Do you happen to have experience with best practices for maintaining a collection of images, and in particular automating rebuilding/tagging/pushing several dependent images?

jeff-cohere: The Docker-related stuff I've done is typically simple enough that I can accomplish rebuilding/tagging/pushing in relatively simple shell scripts. But I've not tried to express dependencies between images in any systematic way. Do you have anything specific in mind? Are you spending a lot of time waiting for certain images to rebuild?

bishtgautam:
Should we now disable Travis-CI as a requirement for PRs?

jeff-cohere:
> Should we now disable Travis-CI as a requirement for PRs?

Sure. If you approve this PR, we can merge it, and I'll apply Jed's speedup suggestions in another one. This removes the Travis workflow files--I can't remember whether we need to do anything else to disable Travis.

bishtgautam (approved):

.github/workflows/auto_pr_test.yml
jedbrown: Sort of. I have one or more containers with mpich and dev tools, then PETSc, then adding other packages. `make` doesn't really track container results (because they aren't local files). I made a half-hearted attempt to use podman and buildah, but they seem to need more configuration to support non-root users and I don't know if it helps for my problem. Ideally there would be a tool that would look at `FROM` directives in Dockerfiles and admit a matrix of `ARG`s to apply.

bishtgautam:
I changed the branch protection rules for master under settings for the repo to exclude Travis-CI.

.github/workflows/auto_pr_test.yml
jeff-cohere: I'll poke around a bit and see if anyone has figured out anything that might help with this.

PR #146, opened 2021-01-07
Improved CMocka demo: pkg-config support, isolated processes

jeff-cohere:
Here's another illustration of how we might use CMocka to write unit tests. This PR uses `pkg-config` to determine whether CMocka is available, and then provides targets for running unit tests. These targets aren't finished yet--they just run all the tests in one process. I wanted to focus for now on how one might poll a test executable for the number of tests and then run them one at a time.

To see all this, try the following:

```
# Build the unit tests. This target also runs the tests in one process, which we don't want, but whatever.
make unit-tests

# Ask our unit test exe how many tests it has (should be 3).
./src/tests/test_tdyinit count

# Try running the tests individually by 0-based index.
./src/tests/test_tdyinit 0
./src/tests/test_tdyinit 1
./src/tests/test_tdyinit 2

# What happens here?
./src/tests/test_tdyinit 3
```

One drawback of this approach is that command line arguments are not available to unit tests--they are interpreted as a user "talking to" the test harness. Arguably, this isn't a drawback, since parameterizing unit tests this way can be hazardous in some circumstances.

If we like this approach, I can wrap this machinery up in a header file that we can just use for our other tests.

### DIFF
--- a/.github/workflows/auto_pr_test.yml
+++ b/.github/workflows/auto_pr_test.yml
@@ -46,6 +46,8 @@ jobs:
         make V=1 transient_mpfaof90 transient_snes_mpfaof90
         cd ../../regression_tests
         num_failures=0
+        make unit-tests
+        num_failures=$(( $num_failures + $? ))
         make test-mpfao
         num_failures=$(( $num_failures + $? ))
         make test-steady
--- a/.github/workflows/auto_pr_test.yml
+++ b/.github/workflows/auto_pr_test.yml
@@ -35,8 +35,10 @@ jobs:
 
     - name: Running tests (${{ matrix.build-type }})
       run: |
-        pkg-config --libs cmocka
         set +e # disable "fast fail" (continue on test failures)
+        num_failures=0
+        make unit-tests
+        num_failures=$(( $num_failures + $? ))
         cd demo/mpfao; make V=1 codecov=1
         cd ../richards; make V=1 codecov=1
         cd ../steady; make V=1 codecov=1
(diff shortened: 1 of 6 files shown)


jeff-cohere:
@jedbrown , have you had a chance to look this over? If we want to adopt this (or a similar) approach that uses CMocka, I can add it to the Docker image that we're now using so it's around for our autotesting. If we think unit testing with MPI is too complicated, maybe we can articulate a different approach to constructing tests that doesn't always rely on "gold" files.

bishtgautam:
I'm taking myself off the reviewer list as I'm not qualified to review it.

jeff-cohere:
This PR is ready for review. I wasn't able to make things "turn-key" for cmocka MPI test programs in the same way as for serial programs, but I've tried to address @jedbrown 's concerns from #139 in a non-intrusive way. it's still true that test output can be confusing in the sense that each process prints it, but addressing that would be a bit more invasive.

jeff-cohere:
Hi @jedbrown and @bishtgautam ,

We've got another PR (#174) that uses this cmocka machinery to illustrate how we can create labels for boundary conditions in meshes. If no one objects, I'd like to get this merged so we can try writing more unit tests and see if e.g. they help us understand some of the PETSc concepts we've been discussing.

bishtgautam (approved):

jedbrown (approved):
Sorry about my extremely delayed review. I think this looks great. If the harness removes the index argument, it should still be possible to pass command line arguments to each test. This is often useful to investigate what goes wrong (e.g., add monitors or adjust the args to reduce). It can be done using the environment variable `PETSC_OPTIONS`, but command line is useful.

If the test fails, can we print the command needed to reproduce without the harness (i.e., with the `mpiexec`) so the user can easily rerun in a debugger?

jeff-cohere:
Thanks, Jed, for approving and for the feedback. Both of those items are doable, and I'll add them in another PR soon.

PR #151, opened 2021-02-04
Fixes the input directories for Doxygen

bishtgautam:
Removes libCEED directories and adds TDycore directories.


jeff-cohere (approved):
Looks good! Sorry about the oversight.

Issue #147, opened 2021-01-29
Refactor TDymesh


bishtgautam:
After some discussion about the difficulty of iterating over various entities within the mesh, we've observed the following things:

* It's difficult to create a "TDyMesh" instance, either by itself from parameters or from a `DMPlex`.
* Looping over the various entities (vertices of a cell, subcells of a vertex, and so on) requires a lot of knowledge of the `TDyMesh` data structure, which makes looping code brittle.

It would be nice if we had a more consistent interface to the mesh, both to simplify its creation (for testing and within the dycore itself) and to access its entities. Here's a proposed interface for a refactored mesh. All functions return the usual error code, with 0 for success and anything else indicating failure.

## Creating and Destroying a Mesh

```
// Create a mesh from a DM object. We need to be careful about specifying preconditions on the DM here.
TDyMeshCreateFromDM(DM dm, TDyMesh* mesh);

// Destructor function for a mesh.
TDyMeshDestroy(TDyMesh* mesh);
```

We might also offer some convenience functions for box meshes, reading from a file, etc.

## Looping Over Entities within a Mesh

```
// Given a mesh and a cell index, retrieve an array of cell vertex indices, and their number.
TDyMeshGetCellVertices(TDyMesh* mesh, PetscInt cell, PetscInt** vertices, PetscInt* num_vertices);

// Given a mesh and a cell index, retrieve an array of indices of faces bounding the cell,
// and their number.
TDyMeshGetCellFaces(TDyMesh* mesh, PetscInt cell, PetscInt** faces, PetscInt* num_faces);

// Given a mesh and a cell index, retrieve an array of indices of neighboring cells, and their number.
TDyMeshGetCellNeighbors(TDyMesh* mesh, PetscInt cell, PetscInt** neighbors, PetscInt* num_neighbors);

// Retrieve the centroid for the given cell in the given mesh.
TDyMeshGetCellCentroid(TDyMesh* mesh, PetscInt cell, TDyCoordinate* centroid);

// Retrieve the volume for the given cell in the given mesh.
TDyMeshGetCellVolume(TDyMesh* mesh, PetscInt cell, PetscReal* volume);

// Given a mesh and a subcell index, retrieve an array of indices of faces bounding the subcell,
// and their number.
TDyMeshGetSubcellFaces(TDyMesh* mesh, PetscInt subcell, PetscInt** faces, PetscInt* num_faces)

// Given a mesh and a vertex index, retrieve an array of subcell indices and their number.
[...]

PR #153, opened 2021-03-02
Added interface functions for efficiently and clearly accessing mesh entities.

jeff-cohere:
See #147 for details on this one. We don't have unit tests yet--a candidate mechanism for that is in PR #146.

There are two things remaining for #147:

1. Self-contained mesh constructor(s) for creating standalone mesh objects.
2. Unit tests that manipulate these objects

### DIFF
--- a/include/private/tdymeshimpl.h
+++ b/include/private/tdymeshimpl.h
@@ -18,14 +18,6 @@ typedef enum {
 } TDySubcellType;
 
 typedef struct {
-  PetscReal X[3];
-} TDyCoordinate;
-
-typedef struct {
-  PetscReal V[3];
-} TDyVector;
-
-typedef struct {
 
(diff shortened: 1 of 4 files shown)


knepley:
Why do you want to do this?

jeff-cohere:
Hey Matt,

These interface routines are an alternative to directly accessing the arrays within the mesh data structure, which doesn't read very easily. Do you think this is a bad idea?

knepley:
I think I understand now. We are already using these structures in the code, and you are just wrapping up the use. That seems reasonable. I would not advocate new code like this, since I would rather we just use the existing API, and fix it if it does not do what we need, but that it too much heavy lifting for this PR,

bishtgautam (approved):

PR #158, opened 2021-04-07
Fixes TPF for parallel runs

bishtgautam:
- Corrects the offset value when converting mesh elements to a compressed format
- Cleans up code by removing unused variables and splitting Gmatrix computation

### DIFF
--- a/src/mpfao/3D/tdympfao3D_core.c
+++ b/src/mpfao/3D/tdympfao3D_core.c
@@ -324,7 +324,6 @@ PetscErrorCode ComputeCandFmatrix(TDy tdy, PetscInt ivertex, PetscInt varID,
   TDyMesh *mesh = tdy->mesh;
   TDyCell *cells = &mesh->cells;
   TDyVertex *vertices = &mesh->vertices;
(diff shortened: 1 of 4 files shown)


src/mpfao/3D/tdympfao3D_core.c
```diff
@@ -38,7 +38,88 @@ PetscErrorCode TDyComputeEntryOfGMatrix3D(PetscReal area, PetscReal n[3],
 }
 
 /* -------------------------------------------------------------------------- */
-PetscErrorCode TDyComputeGMatrixFor3DMesh(TDy tdy) {
+PetscErrorCode TDyComputeGMatrixDefaultFor3DMesh(TDy tdy) {
```
jeff-cohere, line 41: Is it worth documenting how this function differs from `TDyComputeGMatrixFor3DMesh`?

jeff-cohere: Or maybe there's a better name than "Default". Unless everyone already knows that that means. :-)

jeff-cohere (approved):
Looks good! I left a couple very minor comments.

PR #164, opened 2021-04-14
Adds a test for DMPlexExtrude

bishtgautam:
- Adds test for DMPlexExtrude functionality in PETSc
- Fixes wedge mesh and adds a test for it
- Adds few mesh accessor functions


jeff-cohere:
I updated the Docker container to a very recent version of PETSc, which changes the Exodus II viewer interface functions, so you'll probably see some related compilation errors.

demo/transient/transient_snes_mpfaof90.F90
```diff
@@ -217,10 +223,12 @@ program main
     CHKERRA(ierr);
     call DMGetDimension(dm, dim, ierr);
     CHKERRA(ierr);
-    if (dim /= 3) then
-      write(*,*)'Only 3D meshes are supported and the exodus file is not 3D'
-      call exit(0)
-    endif
+    if (dm_plex_extrude_layers > 0) then
+      call DMPlexExtrude(dm, PETSC_DETERMINE, -1.d0, PETSC_TRUE, PETSC_NULL_REAL, PETSC_TRUE, edm, ierr);
```
jeff-cohere, line 227: Just curious--the second argument here is `PETSC_DETERMINE`, which means that PETSc itself extracts the number of layers from `-dm_plex_extrude_layers`. Do we still need the call to `PetscOptionsGetInt` at line 181 above, since we don't actually use the value of `dm_plex_extrude_layers`?

src/mesh/tdymesh.c
```diff
@@ -125,20 +163,29 @@ PetscErrorCode TDyMeshGetVertexInternalCells(TDyMesh *mesh,
   return 0;
 }
 
-/// Given a mesh and a vertex index, retrieve an array of associated subcell
-/// indices and their number.
+/// Given a mesh and a vertex index, return number of associated internal
```
jeff-cohere, line 166: Is an "internal" cell a cell not attached to a domain boundary? Or does this include all cells not attached to parallel subdomain boundaries?

jeff-cohere (approved):
Just a few simple questions. Looks good overall.

src/mesh/tdymesh.c
bishtgautam: "Internal" cell is my lingo and here is the explanation. Boundary faces have an internal cell (DMPlex id>=0) and a fictitious cell (DMPlex id<0).

We probably need to add comments in the code that explicitly states our lingo.

demo/transient/transient_snes_mpfaof90.F90
bishtgautam: At the moment, one needs to explicitly call `DMPlexExtrude`, but @knepley is working on it so we can avoid calling it in the future.

bishtgautam:
let me rebase this branch and resolve conflicts

PR #165, opened 2021-04-19
updated exodus commands with new petsc functions

leorosie:
Updates TDycore so it is compatible with recent changes in PETSc. 
The merge-request that includes changes to Exodus in PETSC 
is https://gitlab.com/petsc/petsc/-/merge_requests/3836/diffs.


bishtgautam (approved):
Thanks for getting this done quickly.

PR #167, opened 2021-04-21
Updates TDyDriver

bishtgautam:
- Adds missing FORTRAN interfaces for existing C functions
- Adds new C functions and their FORTRAN interfaces
- Adds check for TDyDriver to only set default functions for material 
  properties if they haven't been already set.
- Adds a new test


demo/transient/transient_snes_mpfaof90.F90
```diff
@@ -184,6 +185,10 @@ program main
   CHKERRA(ierr);
   call PetscOptionsGetBool(PETSC_NULL_OPTIONS,PETSC_NULL_CHARACTER,"-pflotran_consistent",pflotran_consistent,flg,ierr);
   CHKERRA(ierr)
+
```
jeff-cohere, line 188: Is this a temporary approach for switching this demo to use the driver? Are we planning to start using driver logic exclusively at some point in the future?

src/tdymaterialproperties.c
```diff
@@ -37,26 +43,78 @@ PetscErrorCode MaterialPropertiesDestroy(MaterialProp *matprop){
   PetscFunctionReturn(0);
 }
 
+//
+// Are material properties set
+//
+
+/* -------------------------------------------------------------------------- */
+/// Reports if permeability is set
```
jeff-cohere, line 51: Does this refer to whether a material property has been set either by a command-line argument or by a function call in the program? If so, might be worth mentioning here.

jeff-cohere: Ah, I see, the flags are set only by calls to the appropriate functions.

src/tdycore.c
```diff
@@ -1293,6 +1293,15 @@ PetscErrorCode TDySetDtimeForSNESSolver(TDy tdy, PetscReal dtime) {
   PetscFunctionReturn(0);
 }
 
+PetscErrorCode TDySetInitialCondition(TDy tdy, Vec initial) {
```
jeff-cohere, line 1296: Is it worth describing the form of the solution vector here under various circumstances, since this function is part of the library interface?

jeff-cohere (approved):
Again, just a few minor questions.

demo/transient/transient_snes_mpfaof90.F90
bishtgautam: >Are we planning to start using driver logic exclusively at some point in the future?

You read my mind. I envision switching all 3D+MPFA-O demos to use the driver instead of directly calling `SNESSolve` or `TSSolve`. A new PR will be coming soon.

src/tdycore.c
bishtgautam: Added a comment describing the function

jeff-cohere (approved):

PR #168, opened 2021-04-21
Defines global constant and default parameter values

bishtgautam:
- Fixes the different values of gravity that were being used in the code
- Defines default parameter value in a header file


include/private/tdyconstants.h
```diff
@@ -0,0 +1,18 @@
+#if !defined(TDYCONSTANTS_H)
+#define TDYCONSTANTS_H        
+
+#define GRAVITY_CONSTANT              9.8068
```
jeff-cohere, line 4: The other mechanism that C offers for defining constants like this is the `static const` declaration, which adds type safety but precludes optimizations in some cases (not these):

```
static const PetscReal GRAVITY_CONSTANT = 9.8068;
```

Various people have various opinions about this kind of stuff. I don't think it matters much as long as we are consistent. But the `static const` method does type checking, which can be nice.

jeff-cohere:
Looks like there's a failure with the wedge test. The code changes look fine. I prefer the type safety of `static const` to `#define`, so if no one is opposed to that, maybe we can switch the constants over to that style of declaration.

include/private/tdyconstants.h
jedbrown: I have no objection. `#define GRAVITY_CONSTANT ((PetscReal)9.8068)` is yet another way. In situations like this, there's usually no computational advantage to having it be a compile-time constant, and there are times for debugging that turning off gravity or doubling gravity can be convenient. If constants are controllable via run-time options, then the default constants never need to appear in a header, just in the compilation unit that initializes all the default values.

bishtgautam: @jeff-cohere @jedbrown, I agree that adding the ability to change gravity and other default values from the command line is a good feature. Is this a reasonable approach https://github.com/TDycores-Project/TDycore/pull/168/commits/6a0038922efd26b693c177a5cef712ca5d60a5e3 to avoid defining values as compile-time constants? 

After this discussion, I'm inclined to consolidate the reading of all the command line options in a single function and save them in a new `tdy->options` struct. How does this sound?

jeff-cohere: That sounds good to me, provided that the options are structured in a way that can be understood when we consolidate our demos to a single driver.

src/tdycore.c
```diff
@@ -197,6 +197,18 @@ PetscErrorCode TDySetDM(TDy tdy, DM dm) {
   PetscFunctionReturn(0);
 }
 
+PetscReal TDyGetGravity() {
+  PetscErrorCode ierr;
+  PetscReal gravity = 9.8068;
+
+  ierr = PetscOptionsBegin(PETSC_COMM_WORLD,NULL,"General Options",""); CHKERRQ(ierr);
+  ierr = PetscOptionsReal("-tdy_gravity", "Value of gravity", NULL, gravity, &gravity, NULL); CHKERRQ(ierr);
+  ierr = PetscOptionsEnd(); CHKERRQ(ierr);
```
jedbrown: This is okay, but probably better to collect all constants in a big struct somewhere so the options only need to get parsed once. One other thing I do sometimes is abstract over units, so I write
```c
PetscReal gravity = 9.8068 * meter / (second * second);
```
where `meter` is the size of one meter in internal length units and `second` is the duration of one second in internal time units. Being able to choose characteristic scales like this is useful to scale systems for solvers and time integrator tolerances, and allows one to check for dimensional consistency. In many cases, all user input and output should be converted back to SI units (loses bit-for-bit if you do this with checkpoints), or store the scale factors in the output files so visualization and other human interfaces always see SI.
https://github.com/CEED/libCEED/blob/main/examples/fluids/navierstokes.c#L243

bishtgautam:
This PR is good to go from my end.

src/mpfao/tdympfao.c
```diff
@@ -470,6 +518,14 @@ PetscErrorCode TDyMPFAOInitialize(TDy tdy) {
   if (tdy->ops->computepermeability) {ierr = SetPermeabilityFromFunction(tdy); CHKERRQ(ierr);}
   if (tdy->mode == TH){
     if (tdy->ops->computethermalconductivity) {ierr = SetThermalConductivityFromFunction(tdy); CHKERRQ(ierr);}
+    if (tdy->ops->computesoildensity) {
+      printf("SetSoilDensityFromFunction\n");
```
jeff-cohere, line 522: Do you want to leave this and the other `printf` below in here? Or are these just for debugging?

include/private/tdyoptions.h
```diff
@@ -0,0 +1,24 @@
+#if !defined(TDYOPTIONS_H)
+#define TDYOPTIONS_H
+
+#include <petsc.h>
+
+typedef struct{
+    PetscReal gravity_constant;
+
+    // Default values for material properties
+    PetscReal default_porosity;
```
jeff-cohere, line 10: These aren't really "defaults", are they? It looks like they are initialized to a default and then possibly overridden by command line arguments. If I've understood this properly, maybe we should just get rid of the `default_` prefix.

jeff-cohere: (Looking below) or maybe we mean "default" in the sense of "numeric value" instead of a function that computes it? In that case the `default_` prefix makes more sense.

jeff-cohere (approved):
Looks good to me. I just had a question about some calls to `printf` and about the "default" name. @jedbrown , do you have any other thoughts about this PR?

src/mpfao/tdympfao.c
bishtgautam: done

include/private/tdyoptions.h
bishtgautam: > mean "default" in the sense of "numeric value" instead of a function

Yes, that was my intention.

PR #169, opened 2021-04-27
Fixes a mesh with wedge cells

bishtgautam:
- Fixes the computation of subface area for wedge cell
- Updates prefix of existing command-line options 
- Adds a command-line option to specify an initial condition file and
  modifies an existing test to use the new command-line option


bishtgautam:
This PR will be rebased after #168 is merged.

src/tdycore.c
```diff
@@ -559,6 +563,10 @@ PetscErrorCode TDySetFromOptions(TDy tdy) {
   ierr = PetscOptionsEnd(); CHKERRQ(ierr);
   tdy->setupflags |= TDyOptionsSet;
 
+  if (tdy->init_from_file && tdy->init_with_random_field) {
+    SETERRQ(PETSC_COMM_WORLD,PETSC_ERR_USER,"Both -tdy_init_from_file and -tdy_init_with_random_field cannot be specified");
```
jeff-cohere, line 567: Does `SETERRQ` automatically raise an error, or do we need to call `CHKERRQ` here to get the error message and halt the program?

jeff-cohere (approved):

jeff-cohere:
@bishtgautam, just confirming: this PR is ready to go, right? No longer a WIP?

bishtgautam:
I'm working on adding a test for wedge mesh.

src/tdycore.c
bishtgautam: No need to call `CHKERRQ` see https://www.mcs.anl.gov/petsc/petsc-current/docs/manualpages/Sys/SETERRQ.html

bishtgautam:
@jeff-cohere There is already a test for wedge mesh. So, this is ready to be integrated.

PR #170, opened 2021-04-28
Update some function prototypes and definitions to be consistent about using PetscReal vs. double.

rtmills:
This PR fixes some consistency issues with using `PetscReal` vs. `double` that showed up when I tried building TDycores using single precision (for testing on Intel integrated GPUs that do not support double precision).


bishtgautam (approved):

bishtgautam:
@rtmills When trying to resolve conflicts, I noticed that many demos were still using `double`. I found it easier to do a find-replace in all files than to resolve conflicts. If you are okay with it, we can close this PR and you can review #171.

rtmills:
@bishtgautam, your fix in #171 is definitely better. Let's close this one. (I've already reviewed and approved #171). Thanks!

PR #175, opened 2021-05-17
Fixes the logic to arrange cells in circular order

bishtgautam:
The updated code now puts the cells with minimum numbers of
neighbors at the start and end of array containing the cells
ids that needed to be rearranged. The rearranged array is
next sorted to ensure cell `i` and `i+1` are neighbors.

### DIFF
--- a/src/mesh/tdycoremesh.c
+++ b/src/mesh/tdycoremesh.c
@@ -2285,8 +2285,10 @@ PetscErrorCode ArrangeCellsInCircularOrder(TDy tdy, PetscInt *cellsAbvBlw, Petsc
   // Determine the number of neigbhors
   for (PetscInt ii = 0; ii < ncells; ii++){
     for (PetscInt jj = ii+1; jj < ncells; jj++){
-      nNeighbors[ii] += AreCellsNeighbors(tdy, cellsAbvBlw[ii], cellsAbvBlw[jj]);
-      nNeighbors[jj] += AreCellsNeighbors(tdy, cellsAbvBlw[ii], cellsAbvBlw[jj]);
+      PetscInt result = AreCellsNeighbors(tdy, cellsAbvBlw[ii], cellsAbvBlw[jj]);
+
+      nNeighbors[ii] += result;
+      nNeighbors[jj] += result;
     }
   }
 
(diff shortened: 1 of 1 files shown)


jeff-cohere:
Hey Gautam,

This looks like it does what you say it does. Is this tested by anything that assumes the cells are in such an order? Should we add a test for it, maybe?

bishtgautam:
There is [code](https://github.com/TDycores-Project/TDycore/pull/175/files#diff-2e8de59d7f94a447d9166144c1a228e05325d894e96df2c448c399bbeb52d867R2378), which isn't updated in this PR, that checks to make sure that the sorted cells are such that `i` and `i+1` are neighbors.

src/mesh/tdycoremesh.c
```diff
@@ -2268,14 +2268,92 @@ PetscBool AreCellsNeighbors(TDy tdy, PetscInt cell_id_1, PetscInt cell_id_2) {
...
+
+  PetscInt nNeighbors[ncells];
+  for (PetscInt ii = 0; ii < ncells; ii++) {
+    nNeighbors[ii] = 0;
+  }
+
+  // Determine the number of neigbhors
+  for (PetscInt ii = 0; ii < ncells; ii++){
+    for (PetscInt jj = ii+1; jj < ncells; jj++){
+      nNeighbors[ii] += AreCellsNeighbors(tdy, cellsAbvBlw[ii], cellsAbvBlw[jj]);
```
jeff-cohere, line 2288: If you wanted, you could store the result of `AreCellsNeighbors` here and then add it to both `nNeighbors[ii]` and `nNeighbors[jj]`.

src/mesh/tdycoremesh.c
```diff
@@ -2268,14 +2268,92 @@ PetscBool AreCellsNeighbors(TDy tdy, PetscInt cell_id_1, PetscInt cell_id_2) {
...
+      SETERRQ(PETSC_COMM_WORLD,PETSC_ERR_USER,"Number of cells with minimum number of neighbors is not TWO");
+    }
+
+    // Put the cells with min number of neigbhors at the start and end of the array
+    PetscInt tmpOrder[ncells];
+    count = 0;
+    tmpOrder[0] = cellsAbvBlw[minIdx];
+    for (PetscInt ii = 0; ii < ncells; ii++){
+      if (ii != minIdx && nNeighbors[ii] > minNeighbors) {
+        count = count + 1;
```
jeff-cohere, line 2328: This is trivial, but one of C's idioms (also used in C++) is that `count++` can be used for `count = count + 1`, and `count += n` can be used instead of `count = count + n`. Once you get used to it, it's much more "breezy".

jeff-cohere (approved):
Looks good. I commented on a couple of things. Up to you whether they're worth incorporating.

bishtgautam:
@jeff-cohere I made a new commit to address your comments.

PR #179, opened 2021-05-17
Add saturation in model output

leorosie:


src/tdyio.c
```diff
@@ -42,11 +43,12 @@ PetscErrorCode TDyIOSetMode(TDy tdy, TDyIOFormat format){
   tdy->io->format = format;
   int num_vars = tdy->io->num_vars;
   DM dm = tdy->dm;
-  char *zonalVarNames[1];
+  char *zonalVarNames[num_vars];
 
   PetscInt dim,istart,iend,numCell,numVert,numCorner;
 
   zonalVarNames[0] = tdy->io->zonalVarNames[0];
```
bishtgautam, line 50: How about using a for-loop here to set `zonalVarNames`?

src/tdyio.c
```diff
@@ -73,33 +74,48 @@ PetscErrorCode TDyIOSetMode(TDy tdy, TDyIOFormat format){
...
   if (tdy->io->format == PetscViewerASCIIFormat) {
-    ierr = TDyIOWriteAsciiViewer(v, time);CHKERRQ(ierr);
+    ierr = TDyIOWriteAsciiViewer(p,s,time);CHKERRQ(ierr);
   }
   else if (tdy->io->format == ExodusFormat) {
     char *ofilename = tdy->io->filename;
 
     ierr = TdyIOAddExodusTime(ofilename,time,tdy->io);CHKERRQ(ierr);
-    ierr = TdyIOWriteExodusVar(ofilename,dm,v,tdy->io,time);CHKERRQ(ierr);
+    ierr = TdyIOWriteExodusVar(ofilename,dm,p,s,tdy->io,time);CHKERRQ(ierr);
```
bishtgautam, line 100: Instead of increasing the input argument (`s` for saturation), have two separate calls to the function to write `p` and `s` e.g.


```

ierr = TDyIOWriteExodusVar(filename, dm, P, zonalVarNames[0], tdy->io,time);CHKERRQ(ierr);
ierr = TDyIOWriteExodusVar(filename, dm, S, zonalVarNames[1], tdy->io,time);CHKERRQ(ierr);
...
...

// Update function
PetscErrorCode TDyIOWriteExodusVar(char *ofilename, DM dm, 
  Vec var, char *varname, PetscInt sequence_num, PetscReal time){
...
...
}
```

src/tdyio.c
```diff
@@ -122,24 +138,36 @@ PetscErrorCode TdyIOInitializeHDF5(char *ofilename, DM dm){
...
+  ierr = PetscObjectSetName((PetscObject) P,word);CHKERRQ(ierr);
+  ierr = VecView(P,viewer);CHKERRQ(ierr);
+
+  ierr = PetscObjectSetName((PetscObject) S,word);CHKERRQ(ierr);
+  ierr = DMGetSection(dm, &sec);
+  //Change to field name to Saturation
+  ierr = PetscSectionSetFieldName(sec, 0, "Saturation"); CHKERRQ(ierr);
+  ierr = VecView(S,viewer);CHKERRQ(ierr);
+  //Change back to LiquidPressure
+  ierr = PetscSectionSetFieldName(sec, 0, "LiquidPressure"); CHKERRQ(ierr);
```
bishtgautam, line 164: Why is this switching taking place?

cc: @knepley

bishtgautam (changes requested):

src/tdyio.c
leorosie: We can temporarily change the name of the field in the section so it prints out correctly with VecView() and then change it back after the call to VecView()

I'm not sure if there is a more elegant way to do this though

bishtgautam: Ok. Let's change the function such that we only write one variable at a time with the variable name as an input argument (see [psuedo code](https://github.com/TDycores-Project/TDycore/pull/179/files#r633853680)). This would enable us to just set the fieldname using the variable name.

bishtgautam (commented):
The changes look great!

include/private/tdyioimpl.h
```diff
@@ -7,21 +7,21 @@ struct _p_TDyIO {
...
 PETSC_INTERN PetscErrorCode TDyIOCreate(TDyIO*);
 PETSC_INTERN PetscErrorCode TdyIOInitializeExodus(char*,char*[],DM ,int);
-PETSC_INTERN PetscErrorCode TdyIOAddExodusTime(char*,PetscReal,TDyIO);
-PETSC_INTERN PetscErrorCode TdyIOWriteExodusVar(char*,DM,Vec,TDyIO,PetscReal);
+PETSC_INTERN PetscErrorCode TdyIOAddExodusTime(char*,PetscReal,DM,TDyIO);
+PETSC_INTERN PetscErrorCode TdyIOWriteExodusVar(char*,Vec,char*,TDyIO,PetscReal);
 PETSC_EXTERN PetscErrorCode TdyIOInitializeHDF5(char*,DM);
-PETSC_EXTERN PetscErrorCode TdyIOWriteHDF5Var(char*,Vec,PetscReal);
-PETSC_EXTERN PetscErrorCode TDyIOWriteAsciiViewer(Vec,PetscReal);
+PETSC_EXTERN PetscErrorCode TdyIOWriteHDF5Var(char*,DM,Vec,char*,PetscReal);
```
bishtgautam, line 21: Could also change in this PR functions names from`TdyIO*` to `TDyIO*`?

bishtgautam (approved):
Looks great

PR #184, opened 2021-07-02
Fixes the registry of TDy

bishtgautam:
Correctly registers TDy` when `TDyCreate()` is called without first calling `TDyInit()`.


knepley (approved):
This looks good to me.

PR #185, opened 2021-07-21
Renamed boundary-condition-related functions to better describe their purposes

jeff-cohere:
* TDySetDirichletValueFunction -> TDySetBoundaryPressureFn
* TDySetTemperatureDirichletValueFunction -> TDySetBoundaryTemperatureFn
* TDySetDirichletFluxFunction -> TDySetBoundaryVelocityFn

Boundary conditions are tricky enough! It's important for us to be precise
about our language. C and Fortran interfaces have been updated, and all the
demos use the new functions.

FYI @ghammond86 @nocollier

### DIFF
--- a/demo/SPE10/steady.c
+++ b/demo/SPE10/steady.c
@@ -145,7 +145,7 @@ int main(int argc, char **argv) {
   ierr = TDySetFromOptions(tdy); CHKERRQ(ierr);
 
   ierr = ReadSPE10Permeability(tdy,ang); CHKERRQ(ierr);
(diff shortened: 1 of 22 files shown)


nocollier:
I think this is a good idea and the naming switches seem ok to me. In additional it clears up some ambiguity. In the weak form of the mixed FE formulation the Dirichlet conditions set are velocity conditions and the Neumann are pressures, the reverse of what is typical. That is, a boundary integral terms emerges from integration by parts with pressure leading to a natural means to enforce conditions, but the velocity conditions must be built into the space.

bishtgautam (approved):

PR #189, opened 2021-07-23
Rosie/add input

leorosie:
Adds support for reading initial condition and permeability
dataset from HDF5.


bishtgautam (commented):
This looks great and I following two comments:
- Let's add test(s) for these new features. We should add test(s) that use more than one processor and have spatially heterogeneous values.
- In the past, we haven't been consistent in adding a description of functions. Let's add a description of new functions like https://github.com/TDycores-Project/TDycore/blob/master/src/tdycore.c#L1353

Issue #188, opened 2021-07-23
Should we standardize some input options and add them to TDySetFromOptions?


jeff-cohere:
When I look through some of the demos, it seems to me that they're doing a lot of work that could be collected into a common set of operations that are handled for us, say, by `TDySetFromOptions`. For example, the `transient_snes_mpfaof90` demo has options for

* reading in a mesh file
* extruding the mesh
* setting the type of boundary condition for the domain
* reading initial conditions from a file

It also distributes the mesh and performs a lot of setup that depends on these options. What if we pulled some or all of this stuff into `TDySetFromOptions`?

This wouldn't prevent us from doing things ourselves in a demo--we could poll the list of supplied options, or provide some mechanism for asking the `TDy` object what has or hasn't already been done (such as whether it's read initial conditions or a mesh file). It would slim down these demos a bit and allow us to do more validation for all of these operations.

If we do decide to go this way, it would be useful for us to come up with a standard set of options and acceptable values.

This relates to work I'm doing in #187, in that it would allow me to standardize and simplify the possibilities for reading in meshes and boundary conditions.

jedbrown:
If you're okay to work with PETSc `main`, then all meshes can be read, distributed, and extruded with
```c
DMCreate(comm,&dm);
DMSetType(dm, DMPLEX);
DMSetFromOptions(dm);
```
https://petsc.org/main/docs/manualpages/DM/DMSetFromOptions.html

jeff-cohere:
Just a note, in case we do decide to update to a more recent PETSc: the location of the headers supplying `khash`, which our timers use, has changed. We'll have to make a decision about when we switch and then apply a code change.

bishtgautam:
How about splitting `-tdy_thermal_conductivity <value_or_func>` in two options one for value and one for function?

jeff-cohere:
I've updated the description of this issue to reflect the work that's been done. It'll be easier to finish the rest and clean up our demos once we discuss #200.

PR #192, opened 2021-08-05
Added a function registry for referring to functions by name.

jeff-cohere:
This PR adds a registry for named functions. This lets a demo/driver program register functions (in C or Fortran) that can be selected for boundary conditions. We can also use these functions for forcing functions and other volumetric quantities. All functions live in the same namespace.

The basic functionality is demonstrated in the `transient_snes_mpfaof90` demo. Instead of calling `TDySetBoundaryPressureFn`, we register `PressureFunction` under the name "p0" with a call to `TDyRegisterFunction`, and then we select the boundary pressure function using `TDySelectBoundaryPressureFn` and referring to "p0". The original mechanism (`TDySetBoundaryPressure`) is still supported, and is in fact used by the new registry selection thingy.

This allows us to do all sorts of things that were previously difficult, including using names of functions in command line arguments, and embedding function names in mesh files. We can now register these functions in our demos/drivers and e.g. create mesh files specifically for those drivers that can associate different domain boundary surfaces with different named functions. For now, we only allow a single boundary function, but we can expand that when we're interested in doing so.

(I'm excited by this PR, because I've figured out a way to make the Fortran 2003 interoperability standard cooperate with PETSc's bespoke Fortran bindings!)

Also included: I added a `TDyOnFinalize` function to allow us to register shutdown functions that are called when `TDyFinalize` is called. This allows us to write subsystems without the basic foundation having to know about them.

(In service of #188)


jeff-cohere:
I've added support for selecting porosity and forcing functions using the same method. The `transient_snes_mpfaof90` demo now uses the registry to select its porosity. If we wanted to, we could add command line parameters to this function to tell it which pressure and porosity functions to use.

For permeability (which we're not yet doing in this Fortran demo), we might have to have a "tensor function" Fortran type, to accommodate the different rank of the function's range. I expect we'll be able to discuss this stuff when it becomes more interesting.

demo/transient/transient_snes_mpfaof90.F90
```diff
@@ -305,7 +312,8 @@ program main
   CHKERRA(ierr);
 
   if (pflotran_consistent) then
-     call TDySetPorosityFunction(tdy,PorosityFunctionPFLOTRAN,0,ierr);
+!     call TDySetPorosityFunction(tdy,PorosityFunctionPFLOTRAN,0,ierr);
```
bishtgautam, line 315: Will be eventually make `TDySetPorosityFunction` an internal TDy lib function and not expose it to the world?

jeff-cohere: Sure. If people like the registry idea and want to use it exclusively, I see no reason we couldn't make this and the other similar functions internal/private.

bishtgautam (approved):

PR #193, opened 2021-08-06
Adds cubic smoothing of Mualem relative permeability

bishtgautam:
- Cubic smoothing is added for Mualem relative permeability.
  The approach implemented here follows the [same approach] taken by
  PFLOTRAN.
- New variables are defined so variables are not shared between different relative perm
   and saturation functions
- Few regression files are updated due to the changes caused by cubic interpolation.


include/private/tdycharacteristiccurvesimpl.h
```diff
@@ -18,20 +18,29 @@ typedef struct {
   PetscInt *SatFuncType;         /* type of saturation function */
   PetscInt *RelPermFuncType;     /* type of relative permeability */
   PetscReal *sr;                 /* residual saturation (min) [1] */
-  PetscReal *m;                  /* parameter used in saturation and relative permeability function [-]*/
-  PetscReal *n;                  /* parameter used in Gardner saturation function [-] */
-  PetscReal *alpha;              /* parameter used in VanGenuchten saturation function [-] */
+  PetscReal *gardner_m;          /* parameter used in Garnder saturation [-] */
```
jeff-cohere, line 21: Garnder -> Gardner

src/tdycharacteristiccurves.c
```diff
@@ -212,7 +240,22 @@ void RelativePermeability_Irmay(PetscReal m,PetscReal Se,PetscReal *Kr,
...
+/// @param [inout] *dKr_dSe  derivate of relative permeability w.r.t. Se
+///
+/// kr = Se^{0.5} * [ 1 - (1 - Se^{1/m})^m ]^2     if P < P_ref or Se < 1.0
+///    = 1                                         otherwise
+///
+/// dkr/dSe = 0.5 Se^{-0.5} [ 1 - (1 - Se^{1/m})^m ] +
+///           Se^{0.5}      2 * Se^{1/m - 1/} * (1 - Se^{1/m})^{m - 1} * (1 - (1 - Se^{1/m})^m)  if Se < 1.0
+///         = 0                                                                                  otherwise
+///
+void RelativePermeability_Mualem_NonSmooth(PetscReal m,PetscReal Se,PetscReal *Kr,
```
jeff-cohere, line 258: Maybe `Unsmoothed` instead of `NonSmooth`?

src/tdycharacteristiccurves.c
```diff
@@ -231,6 +274,149 @@ void RelativePermeability_Mualem(PetscReal m,PetscReal Se,PetscReal *Kr,
...
+/// df_dx(low)  = drv_rel_perm_fn(x)  = df1_dx
+/// df_dy(high) = 0.0                 = df2_dx
+///
+/// Linear system:
+/// a0 + a1 * x1 + a2 * x1^2   + a2 * x1^3     = f1
+/// a0 + a1 * x2 + a2 * x1^2   + a2 * x2^3     = f2
+///      a1      + 2 * a2 * x1 + 3 * a3 * x1^2 = df1_dx
+///      a1      + 2 * a2 * x2 + 3 * a3 * x2^2 = df2_dx
+///
+PetscErrorCode CubicPolynomialSetup(PetscReal x1, PetscReal x2, PetscReal *rhs) {
```
jeff-cohere, line 301: If you know that `rhs` should be an array of length 4, you can enforce that in the function declaration with `PetscReal rhs[4]`. That's also a nice cue to users.

src/tdycharacteristiccurves.c
```diff
@@ -231,6 +274,149 @@ void RelativePermeability_Mualem(PetscReal m,PetscReal Se,PetscReal *Kr,
...
+   dgesv_( &n, &nrhs, A, &lda, ipiv, rhs, &ldb, &info );
+
+  PetscFunctionReturn(0);
+}
+
+/* -------------------------------------------------------------------------- */
+///
+/// Computes the value and derivative of the cubic polynomial
+///
+/// @param [in] *coeffs  coefficients of a cubic polynomals
```
jeff-cohere, line 325: We don't normally include the asterisk in documented parameter names. C pointer declarations are weird and confusing--the variable name is `x`, and its type is `PetscReal *`, but folks customarily stick the asterisk next to the variable name to allow the declaration of multiple pointer variables.

For example, suppose you wanted pointers to `PetscReal` values named `x`, `y`, and `z`. One might think that since the type you want is `PetscReal *`, you would say something like

```
PetscReal* x, y, z;
```

This is wrong--this gives you a pointer `x` and two `PetscReal`s `y` and `z`. The right way to declare the three pointers is

```
PetscReal *x, *y, *z;
```

It's one of the most confusing language features.

jeff-cohere: Even more unfortunate is that one "dereferences" a pointer by sticking an asterisk in front of it:

```
*x = 1; // set x to 1
```

So now it's possible for someone to think that `*x` is how you refer to the pointer. The pointer is `x`, and `*x` is the value that it points to. Between that and the type declaration, there's a whole lot of confusion out there.

src/tdycore.c
```diff
@@ -309,11 +310,17 @@ PetscErrorCode TDyMalloc(TDy tdy) {
...
 
+   PetscReal mualem_poly_low = 0.99;
+
   for (c=0; c<nc; c++) {
     cc->sr[c] = options->default_residual_saturation;
-    cc->n[c] = options->default_gardner_n;
-    cc->m[c] = options->default_vangenutchen_m;
-    cc->alpha[c] = options->default_vangenutchen_alpha;
+    cc->gardner_n[c] = options->default_gardner_n;
+    cc->gardner_m[c] = options->default_vangenutchen_m;
```
jeff-cohere, line 318: Looks like VanGenuchten is mis-spelled here.

jeff-cohere (approved):
I found very minor issues with naming and documentation. Approved once fixed.

src/tdycharacteristiccurves.c
```diff
@@ -212,7 +240,22 @@ void RelativePermeability_Irmay(PetscReal m,PetscReal Se,PetscReal *Kr,
...
+/// @param [inout] *dKr_dSe  derivate of relative permeability w.r.t. Se
+///
+/// kr = Se^{0.5} * [ 1 - (1 - Se^{1/m})^m ]^2     if P < P_ref or Se < 1.0
+///    = 1                                         otherwise
+///
+/// dkr/dSe = 0.5 Se^{-0.5} [ 1 - (1 - Se^{1/m})^m ] +
+///           Se^{0.5}      2 * Se^{1/m - 1/} * (1 - Se^{1/m})^{m - 1} * (1 - (1 - Se^{1/m})^m)  if Se < 1.0
+///         = 0                                                                                  otherwise
+///
+void RelativePermeability_Mualem_Unsmooth(PetscReal m,PetscReal Se,PetscReal *Kr,
```
jeff-cohere, line 258: Unsmooth -> Unsmoothed. Sorry for the confusion.

bishtgautam:
@jeff-cohere I added one more commit to address your comments.

jeff-cohere (approved):
Looks great!

PR #194, opened 2021-08-09
Adds new wrapper functions for DM functions

bishtgautam:
TDy wrappers for the following DM functions are added
- Creation of vectors
- Creation of matrix
- Communication of vectors

### DIFF
--- a/src/tdydiscretization.c
+++ b/src/tdydiscretization.c
@@ -130,7 +130,7 @@ PetscErrorCode TDyGlobalToLocal(TDy tdy, Vec global, Vec local){
 
   ierr = DMGetUseNatural(dm, &useNatural); CHKERRQ(ierr);
   if (!useNatural) {
-    PetscPrintf(PETSC_COMM_WORLD,"TDyGlobalToNatural cannot be performed as DMGetUseNatural is false");
+    PetscPrintf(PETSC_COMM_WORLD,"TDyGlobalToLocal cannot be performed as DMGetUseNatural is false");
     exit(0);
   }
 
@@ -146,7 +146,7 @@ PetscErrorCode TDyGlobalToLocal(TDy tdy, Vec global, Vec local){
 /// @param [in] global A PETSc vector
 /// @param [out] local A PETSc vector
 /// @returns 0 on success, or a non-zero error code on failure
-PetscErrorCode TDyNaturalToGlobal(TDy tdy, Vec natural, Vec global){
+PetscErrorCode TDyNaturalToGlobal(TDy tdy, Vec *natural, Vec global){
 
   PetscFunctionBegin;
   DM dm = tdy->dm;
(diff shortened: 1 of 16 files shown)


src/tdydiscretization.c
```diff
@@ -0,0 +1,159 @@
+#include <private/tdycoreimpl.h>
+
+/* -------------------------------------------------------------------------- */
+/// Creates a PETSc global vector
```
jeff-cohere, line 4: Here's an opportunity to include a more precise description of what we mean when we talk about "global", "local", and "natural" vectors. In other words, what does the `i`th degree of freedom in this vector mean? Once we have these documented, it becomes clearer which ones are appropriate to use where.

knepley: Local: A local vector is serial, and lays out data according to the local section. This section assigns dofs to the points of the local Plex.

Global: A global vector is parallel, and lays out data according to the global section. This section assigns dofs to the points of the local Plex, but leaves out any point that is in the pointSF because this means it is not owned by the process.

Natural: A natural vector is a permutation of a global vector, usually to match the input ordering.

jeff-cohere: Right, these are familiar to me. But documenting them with markup can give people some confidence. And because we have a specific family of discretizations, we can be even more specific bout layouts if we want.

jeff-cohere (changes requested):
Looks good! I recommend adding some comments about our definitions of "global", "local", and "natural" as they pertain to the dycore here, as long as we're creating these wrapper functions. But if all these terms are completely obvious to everyone, I won't push on this.

src/tdydiscretization.c
knepley: Ah, good point. Do you want this documentation in PETSc or TDyCore?

src/tdydiscretization.c
```diff
@@ -0,0 +1,159 @@
...
+}
+
+/* -------------------------------------------------------------------------- */
+/// Duplicates a PETSc vector
+///
+/// @param [in] tdy      A TDy struct
+/// @param [in] vector1  A PETSc vector
+/// @param [out] vector2 A PETSc vector
+/// @returns 0 on success, or a non-zero error code on failure
+PetscErrorCode TDyDuplicateVector(TDy tdy, Vec vector1, Vec *vector2){
```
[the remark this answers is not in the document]
knepley: +1

src/tdydiscretization.c
jeff-cohere: I was thinking TDyCore, because we're using these things in specific ways. These functions are already documented in PETSc, right?

knepley: Somewhat yes. Okay, where should I put this documentation do you think?

bishtgautam: I will copy/paste @knepley's definitions in my PR and make a new commit.

jeff-cohere: I was just telling @bishtgautam that I think he should mention some of this stuff in our Doxygen comments for the wrappers he's adding. I don't think you need to do any more documentation unless you think adding more detail to `DMCreate{Global,Local,Natural}Vector` in PETSc would be constructive (which I suppose it could be!). Thanks, though!

src/tdydiscretization.c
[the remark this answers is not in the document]
bishtgautam: Since we now have wrappers for creating various vectors (`TDyCreate*Vector`), I added this new function so we have a  consistent call to TDycore functions instead of PETSc functions directly.

jeff-cohere (approved):

bishtgautam:
@jeff-cohere This is ready for merge.

bishtgautam:
Just realized a review from @knepley is pending.

PR #195, opened 2021-08-10
Consolidating and simplifying command-line options

### LINKED ISSUE #188: Should we standardize some input options and add them to TDySetFromOptions?

jeff-cohere:
This PR gathers our command line options, elaborating on `TDyOptions` and adding features for specifying boundary condition functions and the like. I hope it's easier to understand.

Also, I'm forcing us to be more specific about how we specify a mesh, because our rules to date have been confusing. Here are the new rules:

1. If `TDySetDM` is called before `TDySetFromOptions` in a demo/driver, the DM passed to the dycore in that function is the mesh, period. It can't be overridden. We should only use this function in programs that use a fixed geometry, or that generate their own geometry without leaning on `TDySetFromOptions`.
2. If `TDySetDM` is not called before `TDySetFromOptions`, the user must specify exactly one of `-tdy_read_mesh <filename>` or `-tdy_generate_mesh`. If `-tdy_generate_mesh` is given, the properties of the mesh are extracted from our existing mesh-related command line options (though, as I mention in a comment on this PR, I believe we should eliminate our own options and rely on the options that PETSc provides for generating DMPlexes, for simplicity and to leverage the work of the PETSc team).
3. When `TDySetFromOptions` returns successfully, TDycore has a distributed mesh with the right overlapping cells and all that--there's no need for anything else to be done.

With these rules, it's always clear what the geometry is. This is good, because the word "obvious" is meaningless in every context in which anything matters. :-)

There's still work to be done. PR #188 proposes a lot of new options.


knepley (approved):
This looks simpler and more understandable to me.

jeff-cohere:
> I like the logic also. One question that wasn't clear to me from a quick look over the changes: if TDySetDM is called, and then an option is passed in for a mesh (violating rulle 1), will Options throw an error?

Yes, in the name of specificity. If this is too irritating and we'd rather use the option to override the DM supplied with TDySetDM, that's fine, but this is where I started. If you look at the changes in `tdycore.c`, you can see how these rules are enforced in `TDySetFromOptions` (and also `TDySetDM`).

bishtgautam:
>though, as I mention in a comment on this PR, I believe we should eliminate our own options and rely on the options that PETSc provides for generating DMPlexes, for simplicity and to leverage the work of the PETSc team

What is PETSc's option to read a mesh?

jeff-cohere:
> > though, as I mention in a comment on this PR, I believe we should eliminate our own options and rely on the options that PETSc provides for generating DMPlexes, for simplicity and to leverage the work of the PETSc team
> 
> What is PETSc's option to read a mesh?

If we want to read a mesh, we can use our own `-tdy_read_mesh` option. But if we specify `-tdy_generate_mesh`, we can fall back on the DM and DMPlex-related options supported by `DMSetFromOptions` (see PR #188 for a list)

bishtgautam:
So, instead of options in https://github.com/TDycores-Project/TDycore/blob/master/src/tdydm.c#L19-L53, use the following?
```
-dm_plex_box_dim <dim>	- Set the topological dimension
-dm_plex_box_simplex <bool>	- PETSC_TRUE for simplex elements, PETSC_FALSE for tensor elements
-dm_plex_box_lower <x,y,z>	- Specify lower-left-bottom coordinates for the box
-dm_plex_box_upper <x,y,z>	- Specify upper-right-top coordinates for the box
-dm_plex_box_faces <m,n,p>	- Number of faces in each linear direction
-dm_plex_box_bd <bx,by,bz>	- Specify the DMBoundaryType for each direction
-dm_plex_box_interpolate <bool>	- PETSC_TRUE turns on topological interpolation (creating edges and faces)
```

jeff-cohere:
> So, instead of options in https://github.com/TDycores-Project/TDycore/blob/master/src/tdydm.c#L19-L53, use the following?
> ...

Yes, that's what I'm advocating. The benefit is that recent changes to PETSc (which we now have via the latest update) now make all these options available, and we could grab any new ones that appear without adding our own.

bishtgautam:
@jeff-cohere I agree with your proposal. Let's use PETSc options.

jeff-cohere:
Cool. I'll get rid of the TDycore-specific mesh generation options and retrofit the demos to use the PETSc ones, and I'll let everyone know if I hit any snags.

knepley:
@bishtgautam For reading in a mesh, you can use `-dm_plex_filename`.

include/private/tdymaterialpropertiesimpl.h
```diff
@@ -26,11 +26,11 @@ PETSC_INTERN PetscBool TDyIsThermalConductivytSet(TDy);
...
-PETSC_INTERN PetscErrorCode TDySoilDensityFunctionDefault(TDy,PetscReal*,PetscReal*,void*);
-PETSC_INTERN PetscErrorCode TDySoilSpecificHeatFunctionDefault(TDy,PetscReal*,PetscReal*,void*);
-PETSC_INTERN PetscErrorCode TDyPermeabilityFunctionDefault(TDy,PetscReal *,PetscReal *,void*);
-PETSC_INTERN PetscErrorCode TDyThermalConductivityFunctionDefault(TDy,PetscReal *,PetscReal *,void*);
-PETSC_INTERN PetscErrorCode TDyPorosityFunctionDefault(TDy,PetscReal *,PetscReal *,void*);
+PETSC_INTERN PetscErrorCode TDyConstantSoilDensityFunction(TDy,PetscReal*,PetscReal*,void*);
+PETSC_INTERN PetscErrorCode TDyConstantSoilSpecificHeatFunction(TDy,PetscReal*,PetscReal*,void*);
+PETSC_INTERN PetscErrorCode TDyConstantPermeabilityFunction(TDy,PetscReal *,PetscReal *,void*);
+PETSC_INTERN PetscErrorCode TDyConstantThermalConductivityFunction(TDy,PetscReal *,PetscReal *,void*);
+PETSC_INTERN PetscErrorCode TDyConstantPorosityFunction(TDy,PetscReal *,PetscReal *,void*);
```
jedbrown, line 33: Not pertinent to the name change, but we may want to make these vectorizable, taking arguments `(TDy, PetscInt, const PetscReal *, PetscReal *, void *)`.

src/tdyregression.c
```diff
@@ -201,7 +201,7 @@ PetscErrorCode TDyRegressionOutput(TDy tdy, Vec U) {
   ierr = VecRestoreArray(U,&u_p); CHKERRQ(ierr);
   ierr = VecRestoreArray(U_pres,&pres_p); CHKERRQ(ierr);
 
-  if (tdy->mode == TH) {
+  if (tdy->options.mode == TH) {
```
jedbrown, line 204: Also not pertinent to this step, but it would be nice to use callbacks instead of if statements, which will become `switch` statements if there are ever more than two modes.

jedbrown (approved):

include/private/tdymaterialpropertiesimpl.h
jeff-cohere: That's a good idea and one we should adopt.

src/tdyregression.c
jeff-cohere: Can you give an example of what you mean by this?

jeff-cohere:
I don't think this covers all the ground mentioned in #188, but things are working, and I kinda don't want to make this PR much bigger. @bishtgautam , let me know if you'd like me to address anything else. @jedbrown , I made an issue for "vectorizing" our functions.

src/tdyregression.c
jedbrown: For example, the mode could have a `tdy->diagnostics(tdy, ...)` that computes/reports any diagnostics specific to that mode. This saves having a sequence of switch statements with possibly fragile carried dependencies.

bishtgautam (approved):

src/tdyregression.c
jeff-cohere: Ah, okay.  A "virtual table". Yes, that makes sense.

Issue #200, opened 2021-08-13
Pruning TDycore


jeff-cohere:
In the interest of focusing on what we need to be relevant to a project like E3SM, I think it would be good for us to start a conversation about what we can get rid of, now that we know a bit more what we're doing.

Here are some starter questions (at varying levels of provocation):

* What algorithms are we no longer including in the TDycore library?
* What demos are no longer relevant?
* What are the combinations of modes/discretizations/solvers that we wish to support explicitly?
* Do we still need 2D capability (e.g. for transect calculations), or can we get everything we need from 3D?

@bishtgautam , if it's appropriate, maybe we can talk about this at our next joint meeting.

bishtgautam:
### **Fate of demos**

1. `SPE10/steady.c`
   - No test
2. `mpfao/mpfao.c`
    - Steady state MPFA-O tests (mostly 2D, one 3D test, one test using an exodus file 64xy_3z_wedge.exo)
    - **Delete the demo and move the 64xy_3z_wedge.exo test in** `richards/richards_driver.c`
3. `qa_test/qa_test.c`
   - For QAQC framework
   - **Keep it for now**
4. `richards/richards_driver.c`
    - MPFA-O tests for SNES/TS, serial/parallel, reading permeability datasets
    - **DEFINITELY keep this demo**
5. `steady/steady.c`
   - MFE tests: WY, TPF and BDM
6. `steadyblock/steadyblock.c`
   - No test
   - **Delete the demo**
7. `steadyf90/steadyf90.F90`
    - This is a FORTRAN version of `steady/steady.c`
    - **Delete the test**
8. `th/th_driver.c`
    - Serial and parallel tests for TH
    - **DEFINITELY keep this demo**
9. `transient/transient.c`
   - WY serial and parallel tests
10. `transient/transient_mpfao.c`
    - MPFA-O tests for various time integrators: TS, TS+DAE, TS_TRANSIENTVAR
    - **Delete the demo and move the tests in** `richards/richards_driver.c`
11. `transient/transient_mpfaof90.F90`
    - MPFA-O tests using TS time-integrators for different 
       1. `discretization methods (MPFA-O/2 point flux)
       2. `different BCs
    - **Keep the demo**
12. `transient/transient_snes_mpfaof90.F90`
    - MPFA-O tests using SNES time-integrators for multiple configurations
    - **DEFINITELY keep this demo**
13. `transient/transient_th_mpfao.c`
    - Tests transient TH
[...]

PR #204, opened 2021-08-18
Removes 2D MPFA-O code and tests

### LINKED ISSUE #46: Remove 2D MPFA-O code
It doesn't seem worthwhile to keep a separate 2D code when a 2D problem can be formulated as a 3D problem with  number of grid cells in one of the dimensions is one.

cc: @ghammond86

jeff-cohere:
This PR gets rid of all 2D logic in the MPFA-O code and merges the `src/mpfao/3D` source directory into `src/mpfao`. Related functions with `3D` in their names are renamed to simpler names. All 2D MPFA-O tests have been removed. This is the first step of the pruning process outlined in #200.

Closes #46


demo/mpfao/mpfao.cfg
```diff
@@ -1,41 +1,18 @@
...
-	mpfao-prob4
 	mpfao-3d-prob
 #	mpfao-64xy-3z-wedge
 
-standard_parallel=
-	mpfao-prob3-np3
-	mpfao-prob3-perturb-np3
+#standard_parallel=
+#	mpfao-prob3-np3
+#	mpfao-prob3-perturb-np3
```
bishtgautam, line 8: Delete these lines, as the tests have been deleted

jeff-cohere:
Looks like we're losing some code coverage here. I guess that's not too surprising, considering the tests we've jettisoned. @bishtgautam , let me know if you're concerned about this in the short term. Further pruning might improve coverage.

demo/steady/steady.cfg
```diff
@@ -16,7 +16,7 @@ standard=
 	steady-bdm-prob3
 	steady-bdm-prob4
 
-standard_parallel=steady-wy-np3
+#standard_parallel=steady-wy-np3
```
bishtgautam, line 19: same comment as the previous one.

bishtgautam (approved):
I have two minor comments

bishtgautam:
I'm not worried about the decrease in code coverage. At some point, we will have to add tests to improve code coverage.

PR #205, opened 2021-08-19
Pruning demos

jeff-cohere:
This PR deletes some demos and moves some tests around as suggested by @bishtgautam in #200.


jeff-cohere:
That test coverage report is probably underreporting coverage, because it was generated by a configuration in which several tests weren't run.

bishtgautam (approved):
👍

PR #208, opened 2021-09-10
Adds creation of a vector

bishtgautam:
Adds the missing call to create a PETSc vector.


jeff-cohere (approved):
This code must not have been working before. Do we have any I/O tests?

bishtgautam:
There is the following test for checkpoint

https://github.com/TDycores-Project/TDycore/blob/master/demo/richards/richards.cfg#L51

On my local machine, I do get an error, but the automated testing is not reporting an error.

PR #209, opened 2021-09-10
[WIP] Save information about mesh partition for debugging

bishtgautam:
Each rank independently outputs binary vectors for control volumes and vertices.

- The vector for cells has the following information:
  1. Natural ID
  2. Is the cell a local cell (=1) or ghost cell (=0)
  3. Number of vertices forming the cell
  4. IDs of vertices (in local order)

- The vector for vertices has the following information
  1. x,y,z coordinate
  2. Is vertex a local vertex (=1) or ghost vertex (=0) 
  3. Number of cells that share the vertex
  4. IDs of cells (in local order)

### DIFF
--- a/src/tdyutils.c
+++ b/src/tdyutils.c
@@ -482,6 +482,31 @@ PetscInt TDyReturnIndexInList(PetscInt *list, PetscInt nlist, PetscInt value) {
 }
 
 /* -------------------------------------------------------------------------- */
+/// Outputs a sequential PETSc Vec as binary
+///
+/// @param [in] vec A sequential PETSc Vec
+/// @param [in] filename Name of the file
+/// @returns 0 on success, or a non-zero error code on failure
+PetscErrorCode TDySavePetscVecSeqAsBinary(Vec vec, const char filename[]) {
+
+  PetscViewer viewer;
+  PetscErrorCode ierr;
+
+  PetscFunctionBegin;
+
+  ierr = PetscViewerBinaryOpen(PETSC_COMM_SELF, filename, FILE_MODE_WRITE, &viewer); CHKERRQ(ierr);
+  ierr = VecView(vec, viewer); CHKERRQ(ierr);
+  ierr = PetscViewerDestroy(&viewer); CHKERRQ(ierr);
+
+  PetscFunctionReturn(0);
+}
+
+/* -------------------------------------------------------------------------- */
+/// Outputs a PETSc Vec as binary
+///
+/// @param [in] vec A PETSc Vec
+/// @param [in] filename Name of the file
+/// @returns 0 on success, or a non-zero error code on failure
 PetscErrorCode TDySavePetscVecAsBinary(Vec vec, const char filename[]) {
 
   PetscViewer viewer;
@@ -498,7 +523,7 @@ PetscErrorCode TDySavePetscVecAsBinary(Vec vec, const char filename[]) {
 }
 
 /* -------------------------------------------------------------------------- */
-/// Outputs a PETSc Vec as binary
+/// Reads a PETSc Vec as binary
 ///
 /// @param [in] vec A PETSc Vec
 /// @param [in] filename Name of the file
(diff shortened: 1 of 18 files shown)


jeff-cohere:
Hey @bishtgautam . Do you think we should stick the related command-line option into one of our tests to get test coverage on this new feature?

src/tdyutils.c
```diff
@@ -484,6 +484,31 @@ PetscInt TDyReturnIndexInList(PetscInt *list, PetscInt nlist, PetscInt value) {
 }
 
 /* -------------------------------------------------------------------------- */
+/// Outputs a sequential PETSc Vec as binary
+///
+/// @param [in] vec A sequential PETSc Vec
+/// @param [in] filename Name of the file
+/// @returns 0 on success, or a non-zero error code on failure
+PetscErrorCode TDySavePetscVecSeqAsBinary(Vec vec, const char filename[]) {
```
jeff-cohere, line 492: I was going to say that I didn't think it was necessary to have a separate function for outputing a sequential  vector this way, but I can't find any way to get the communicator for a Vec object from the PETSc interface. So I guess we do need it!

jeff-cohere (approved):
Looks good to me. I think it would be a good idea to make sure one of our tests uses this feature, just to make sure it does what we think it does.

src/tdyutils.c
jedbrown: https://petsc.org/release/docs/manualpages/Sys/PetscObjectGetComm.html

This is also powerful, enabling `-some_prefix_vec_view vtk:that-vec.vtu` and lots of other kinds of viewing.
https://petsc.org/release/docs/manualpages/Sys/PetscObjectViewFromOptions.html

jeff-cohere: Of course. Thanks, Jed. I had not thought to look "up the chain." I've been doing too much C++ lately.

@bishtgautam , in that case, we don't need an additional function for writing out vectors with binary viewers. We can just use the `PetscObjectGetComm` function to retrieve the MPI communicator in the existing function, and use that to open the viewer. It should be `PETSC_COMM_SELF` in the case of a serial vector.

And it's worth considering whether to just rely on `PetscObjectViewFromOptions` for viewing. It seems to offer a lot of flexibility.

jedbrown: Note that `VecView(X, NULL)` uses stdout on a matching communicator. Useful for debugging, or `call VecView(X, 0)` in gdb.

knepley: I advocate the `ViewFromOptions` strategy since it is so extensible

bishtgautam: @knepley @jedbrown , I'm trying to understand `VecViewFromOptions` in PETSc `src/snes/tutorials/ex12.c`

- `ierr = VecViewFromOptions(u, NULL, "-guess_vec_view");CHKERRQ(ierr);` is at [ex12.c#L862](https://gitlab.com/petsc/petsc/-/blob/main/src/snes/tutorials/ex12.c#L862)
- Then looking at a test on [ex12.c#L1636](https://gitlab.com/petsc/petsc/-/blob/main/src/snes/tutorials/ex12.c#L1636) that includes `-guess_vec_view glvis:`

By specifying `-guess_vec_view glvis:`, is the Vec being written out in a `glvis` format? Is the output displayed on the screen because there isn't a string specified after `:` for the output to be saved to a file?

jedbrown: Yeah, I think you'd need a glvis server running (to receive and render this data). The option is powerful in providing many different ways to view.

jeff-cohere:
Hey @bishtgautam . Is this ready for (re-)review? I see a lot of time-integrator-related changes in addition to the debugging information you describe above. Are those from another branch that got merged, or just stuff you're including in this PR?

bishtgautam:
@jeff-cohere It is still a WIP and I updated the title to reflect it. Thanks for checking in about this PR.

bishtgautam:
Based on our discussion during today's call, I'm closing this PR. These mesh diagnostics functions will be re-implemented and outputted in HDF5 format.

PR #214, opened 2021-10-18
TDycore polymorphism, part 2

jeff-cohere:
This is the second in a series of "surgical" pull requests to move TDycore to a polymorphic structure like the one used in PETSc (#197).


bishtgautam (approved):
This is great work :+1 Looking forward to the next PR that cleans up MPFA-O stuff

jeff-cohere:
@jedbrown , sorry for the poke (I'm sure you're pretty busy these days)! Do you want to weigh in on this one, or are you okay with TDycore hashing out the details as long as we end up using the DM as the source of truth in the way you're shooting for with libceed? No hurry if you want to take a look--I value your input.

PR #215, opened 2021-11-15
TDycore polymorphism, part 3: separating concerns

### LINKED ISSUE #22: Consolidate data structures between MFE and MPFA

### LINKED ISSUE #133: Warnings in WY
There are few warnings in the WY code:
```
/Users/bish218/projects/tdycore/TDycore-master/src/tdywy.c: In function 'IntegrateOnFace':
/Users/bish218/projects/tdycore/TDycore-master/src/tdywy.c:387:13: warning: variable 'xq' set but not used [-Wunused-but-set-variable]
   PetscReal xq[3],x[27],J[9],N[24],DF[81],DFinv[81],value;
             ^~
/Users/bish218/projects/tdycore/TDycore-master/src/tdywy.c: In function 'TDyWYVelocityNorm':
/Users/bish218/projects/tdycore/TDycore-master/src/tdywy.c:842:13: warning: variable 'xq' set but not used [-Wunused-but-set-variable]
   PetscReal xq[3],x[100],DF[100],DFinv[100],J[100],N[24],vel[3],ve,va,flux0,flux,
             ^~
```

### LINKED ISSUE #196: Make material property, boundary condition functions vectorizable
@jedbrown has suggested in #195 that we modify the signature of the functions we have been using for material properties and boundary conditions to accept an array of input positions and populate an array of outputs. Specifically, we want to change from a "single-point evaluation" approach, e.g.

```
f(tdy, x, v, ctx)
```

to a multi-point evaluation approach:

```
f(tdy, n, X, V, ctx)
```

In the single-point approach,

* `f` is one of these functions
* `tdy` is a `TDy` object passed as the first argument
* `x` is a pointer to a 2- or 3-tuple of real numbers identifying a point in 2D or 3D space
* `v` is a pointer to a real value that stores the (scalar) value of `f`
* `ctx` is an opaque pointer that stores any additional contextual information needed by `f`

The multi-point approach replaces `x` and `v` with

* `n`, an integer defining the number of points at which to evaluate `f`
* `X`, an array of `2*n` or `3*n` values defining `n` 2D or 3D points, with components stored adjacently
* `V`, an array of `n` values that store each of the `n` values of `f` at the points given by `X`

This is a pretty easy thing to do and will probably buy us some performance.

### LINKED ISSUE #197: Rework TDycore with polymorphic "ops" tables

jeff-cohere:
Continuing in our journey (see #211 and #214):

This pull request defines interfaces for all important operations needed by our different numerical discretizations. Specifically:

* `TDy` is now just a container with all the relevant information that describes a problem
* The `TDy` struct contains a context pointer that is used to manage the specific implementations using the function pointers defined in the `ops` table.
* The context pointer stores the WY, BDM, or MPFA-O implementation selected by `TDySetDiscretization`.
* The calculation of boundary conditions (for pressure, temperature, velocity) and sources/sinks (forcings) have been separated into a `Conditions` type. They have been "vectorized" in the sense that they now can compute their values on multiple points.
* All model data has been removed from the `MaterialProps` and `CharacteristicCurves` types, which now have reworked, vectorized interfaces. This allows different implementations to store material properties differently, instead of forcing them to have similar layouts. The `MaterialProps` and `CharacteristicCurves` interfaces now only compute model data.
* Moving model data out of the above types required some changes in the IO system, in addition to those made in the specific numerical implementations.
* There's a new `EOS` type that calculates properties of water.
* Some source files have been moved into new directories:
    * `fe`: the finite element discretizations and related utilities live here
    * `materials`: `EOS`, `CharacteristicCurves`, and `MaterialProp` live here
    * `mpfao`: the MPFA-O code lives here (not a new folder, but it has taken on new code)
* Many functions have been removed from the public interface and made internal.
* The MPFA-O mesh has been consolidated into a smaller number of source files, and it has been separated from TDy itself (because it's part of the MPFA-O implementation).
* The Fortran interface has been reworked with vectorized functions for material properties, boundary conditions, and forcings, like its C counterpart. All Fortran-related source files now live in `src/f90-mod`.
* Several memory leaks in the MPFA-O code have been plugged.
* **Bonus**: source files in the library are now built in a deterministic order, which helps when you're trying to fix a lot of things in various places. :-)

In particular, there's a much stronger sense of an "interface" for TDycore. Instead of any piece doing whatever it wants at any point, we're moving toward a set of interfaces that perform specific tasks in specific contexts in a way that suggests a flow of information that is easier to understand.

## So where are we?

Everything works.

* I had to disable a test and some diagnostic output in `transient_snes_mpfaof90` because it uses some low-level functions that arguably don't belong in a demo, like directly setting coefficients and extracting data, which implies knowledge of how things are indexed. Maybe we should discuss this when we have a chance.
* I've disabled the mesh geometry reading/writing capability, since it makes assumptions about the mesh construction process that no longer hold.
* The TH demos exhibited a very tricky problem that led to the addition of a `set_dm_fields` operation, required by all models, that set up the field layout on a DM before it's distributed. The way it was done before was not correct, but for some reason, we hadn't experienced any problems with it till now.

## What's next?

The bad news is: not everything will be perfect after this PR goes in. A lot of cleanup remains, but much of it is within the separate numerical implementations, which lets different parts of our team adjust things without affecting others.

The good news: the changes in this PR make it easier for us to move in the direction of the DM-oriented problem definition that we started discussing in #211.  Briefly: we attach functions to a DM for computing residuals, Jacobians, etc, and then we hand the DM over to the solvers. This allows us to make TDycore's interface smaller and tighter--the dycore produces a DM that gets used by the solvers directly.

Fixes #133
Fixes #196
Fixes #197
Fixes #22


jeff-cohere:
I've identified a discrepancy in the calculation of T*P in a failing demo test.

jeff-cohere:
Okay, I've fixed some issues I found in the higher-level driver interface, which drives the `richards` and `th` demos. Now these drivers run, but they get bad answers. I could go ahead and find out exactly what's going on, but it would be faster if someone who knew more about the driver interface could help me understand the assumptions that it makes.

These demos and the transfer of saved geometry info in the `transient_snes_mpfaof90` demo are the only ones remaining, as far as I can tell. @bishtgautam , maybe we should discuss how to proceed.

jeff-cohere:
The Richards driver tests now pass. The TH ones still fail.

Also, I found the source of the issue with reading in mesh geometry from a binary file: the new mesh construction process has already converted indices to compressed row format by the time the data is read, whereas the old construction process read in the geometry before doing that conversion.

jeff-cohere:
@bishtgautam and I agreed that we can comment out the mesh geometry reading/writing stuff for now, as we may not actually need it.

The TH tests are the only remaining failures.

src/materials/tdycharacteristiccurves.c
```diff
@@ -0,0 +1,672 @@
...
+        PetscReal m = sat->parameters[type][3*i1+1];
+        PetscReal alpha = sat->parameters[type][3*i1+2];
+        PressureSaturation_Gardner(n, m, alpha, Sr[i], Pc[i],
+                                   &(S[i]), &(dSdP[i]), &(d2SdP2[i]));
+      } else if (type == SAT_FUNC_VAN_GENUCHTEN) {
+        PetscReal m = sat->parameters[type][2*i1];
+        PetscReal alpha = sat->parameters[type][2*i1+1];
+        PressureSaturation_VanGenuchten(m, alpha, Sr[i], Pc[i],
+                                        &(S[i]), &(dSdP[i]), &(d2SdP2[i]));
+      }
```
bishtgautam, line 225: similarly, add `else`

src/materials/tdycharacteristiccurves.c
```diff
@@ -0,0 +1,672 @@
...
+        RelativePermeability_Irmay(m, Se[i], &(Kr[i]), &(dKrdSe[i]));
+      }
+    } else if (type == REL_PERM_FUNC_MUALEM) {
+      for (PetscInt i = 0; i < num_points; ++i) {
+        PetscReal m = rel_perm->parameters[type][6*i];
+        PetscReal poly_low = rel_perm->parameters[type][6*i+1]; // cubic interp cutoff
+        PetscReal *poly_coeffs = &(rel_perm->parameters[type][6*i+2]); // interp coeffs
+        RelativePermeability_Mualem(m, poly_low, poly_coeffs, Se[i], &(Kr[i]),
+                                    &(dKrdSe[i]));
+      }
```
bishtgautam, line 337: add `else`

demo/transient/transient_snes_mpfaof90.F90
```diff
@@ -295,73 +297,53 @@ program main
...
+!        index(c) = c-1;
+!        call MaterialPropAlpha_PFLOTRAN(alpha(c))
+!        call MaterialPropM_PFLOTRAN(m(c))
+!     enddo
+!
+!     call TDySetCharacteristicCurveVanGenuchtenValuesLocal(tdy,ncell,index,m,alpha,ierr)
+!     CHKERRA(ierr);
+!
+!     call TDySetCharacteristicCurveMualemValuesLocal(tdy,ncell,index,m,ierr)
+!     CHKERRA(ierr);
```
bishtgautam, line 339: Is there a new issue to add this functionality back in the future?

demo/transient/transient_snes_mpfaof90.F90
```diff
@@ -417,12 +399,13 @@ program main
...
     CHKERRA(ierr);
 
-    call TDyGetLiquidMassValuesLocal(tdy,nvalues,liquid_mass,ierr)
-    CHKERRA(ierr);
-    mass_pre = 0.d0
-    do g = 1,nvalues
-    mass_pre = mass_pre + liquid_mass(g)
-    enddo
+    ! TODO: Mass conservation diagnostic output is disabled for the moment.
+    !call TDyGetLiquidMassValuesLocal(tdy,nvalues,liquid_mass,ierr)
```
bishtgautam, line 403: Is there a new issue to add this functionality back in the future?

demo/transient/transient_snes_mpfaof90.F90
```diff
@@ -468,8 +452,8 @@ program main
 
   end do
 
-  call TDyGetSaturationValuesLocal(tdy,nvalues,liquid_sat,ierr)
-  CHKERRA(ierr);
+  !call TDyGetSaturationValuesLocal(tdy,nvalues,liquid_sat,ierr)
```
bishtgautam, line 455: Is there a new issue to add this functionality back in the future?

include/private/tdymeshimpl.h
```diff
@@ -170,18 +179,114 @@ typedef struct TDyMesh {
...
 } TDyMesh;
 
-PETSC_INTERN PetscErrorCode TDyBuildMesh(TDy);
-PETSC_INTERN PetscErrorCode TDyAllocateMemoryForMesh(TDy);
+PETSC_INTERN PetscErrorCode TDyMeshCreate(DM,PetscReal*,PetscReal*,PetscReal*,TDyMesh**);
+PETSC_INTERN PetscErrorCode TDyMeshDestroy(TDyMesh*);
+
+// These don't work, and we'll likely get rid of them.
+//PETSC_INTERN PetscErrorCode TDyMeshReadGeometry(TDyMesh*,const char*);
+//PETSC_INTERN PetscErrorCode TDyMeshWriteGeometry(TDyMesh*,const char*);
```
bishtgautam, line 203: Is there a new issue to add this functionality back in the future?

demo/transient/transient_snes_mpfaof90.F90
jeff-cohere: Yes--see #216

demo/transient/transient_snes_mpfaof90.F90
jeff-cohere: Not yet. I think we should discuss how we'd like to do this and the similar stuff below. We need to be careful about what we guarantee to be true when we don't necessarily know the discretization (from outside the library). Alternatively, we could have functions that begin with `TDyMPFAO` for this, which would tie the discretization to the finite volume one, for example.

src/materials/tdycharacteristiccurves.c
```diff
@@ -0,0 +1,672 @@
...
+    } else if (type == SAT_FUNC_VAN_GENUCHTEN) {
+      PetscInt num_points = sat->num_points[type];
+      for (PetscInt i = 0; i < num_points; ++i) {
+        PetscInt j = sat->points[type][i];
+        PetscReal m = sat->parameters[type][2*i];
+        PetscReal alpha = sat->parameters[type][2*i+1];
+        PressureSaturation_VanGenuchten(m, alpha, Sr[j], Pc[j], &(S[j]), &(dSdP[j]),
+                                        &(d2SdP2[j]));
+      }
```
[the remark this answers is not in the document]
jeff-cohere: Good idea. I've patched those in.

demo/transient/transient_snes_mpfaof90.F90
jeff-cohere: Yep. I've added #217

include/private/tdymeshimpl.h
jeff-cohere: It depends on whether we need it. Is this just for saving time in debugging? At this point, we could make a proper effort to optimize the calculation of geometry so we don't need this feature.

demo/transient/transient_snes_mpfaof90.F90
jeff-cohere: See #217

bishtgautam (approved):

PR #220, opened 2022-01-14
Added TDyCreateDiagnostics and TDyComputeDiagnostics functions.

### LINKED ISSUE #217: Post-215 work: add support for retrieving saturation values directly from a model
These functions have been discarded as part of #215, since they're specific to the discretization. We can resurrect them in a more specific form if we can figure out how to articulate what kind of space these values live in (e.g. a finite element function space, or "on cells" in a finite volume method).

```
TDySetCharacteristicCurveVanGenuchtenValuesLocal
TDySetCharacteristicCurveMualemValuesLocal
```

jeff-cohere:
This PR implements the solution to obtaining "diagnostic" fields computed from the solution vector, as discussed in #217. These functions are part of the TDycore interface and have their own implementation-specific function pointers. Currently, they are only implemented for the MPFA-O method.

See `demo/transient/transient_snes_mpfaof90.F90` for how these functions are used in practice.

Fixes #217

### DIFF
--- a/demo/transient/transient_snes_mpfaof90.F90
+++ b/demo/transient/transient_snes_mpfaof90.F90
@@ -180,8 +180,11 @@ program main
 implicit none
 
   TDy                 :: tdy
-  DM                  :: dm
-  Vec                 :: U
+  DM                  :: dm, diags_dm
+  Vec                 :: U, diags
+  PetscSection        :: diags_section
+  PetscInt            :: n_diags, i_diag, i_liquid_mass
+  character (len=256) :: diag_name
   !TS                 :: ts
   SNES                :: snes
   PetscInt            :: rank, successful_exit_code
@@ -376,6 +379,24 @@ implicit none
   call TDySetInitialCondition(tdy,U,ierr);
   CHKERRA(ierr);
 
+  ! Initialize the diagnostics DM and Vec, fetch the section that holds the
+  ! layout, and determine which diagnostic index corresponds to the liquid
+  ! mass (to check mass conservation).
+  call TDyCreateDiagnostics(tdy, diags_dm, ierr)
+  CHKERRA(ierr)
+  call DMCreateGlobalVector(diags_dm, diags, ierr)
+  CHKERRA(ierr)
+  call DMGetLocalSection(diags_dm, diags_section, ierr)
+  CHKERRA(ierr)
+  call PetscSectionGetNumFields(diags_section, n_diags, ierr)
+  do i_diag = 1,n_diags
+    call PetscSectionGetFieldName(diags_section, i_diag, diag_name, ierr)
+    if (diag_name == "liquid_mass") then
+      i_liquid_mass = i_diag
+    end if
+  end do
+
+  ! Set up the SNES solver.
   call TDySetPreviousSolutionForSNESSolver(tdy, U, ierr)
   CHKERRA(ierr);
 
@@ -399,7 +420,12 @@ implicit none
     call TDyPreSolveSNESSolver(tdy,ierr);
     CHKERRA(ierr);
 
-    ! TODO: Mass conservation diagnostic output is disabled for the moment.
+    ! Check the total liquid mass before the step. Because the liquid mass is
+    ! nonnegative, the sum over all cells is the same as the 1-norm.
+    call TDyComputeDiagnostics(tdy, diags_dm, diags, ierr)
+    CHKERRA(ierr)
+    call VecStrideNorm(diags, i_liquid_mass, NORM_1, mass_pre, ierr)
+    CHKERRA(ierr)
     !call TDyGetLiquidMassValuesLocal(tdy,nvalues,liquid_mass,ierr)
     !CHKERRA(ierr);
     !mass_pre = 0.d0
@@ -428,14 +454,19 @@ implicit none
     call TDyPostSolveSNESSolver(tdy,U,ierr);
     CHKERRA(ierr);
 
-    ! TODO: Mass conservation diagnostic output is disabled for the moment.
+    ! Check the total liquid mass after the step. Because the liquid mass is
+    ! nonnegative, the sum over all cells is the same as the 1-norm.
+    call TDyComputeDiagnostics(tdy, diags_dm, diags, ierr)
+    CHKERRA(ierr)
+    call VecStrideNorm(diags, i_liquid_mass, NORM_1, mass_post, ierr)
+    CHKERRA(ierr)
     !call TDyGetLiquidMassValuesLocal(tdy,nvalues,liquid_mass,ierr)
     !CHKERRA(ierr);
     !mass_post = 0.d0
     !do g = 1,nvalues
     !mass_post = mass_post + liquid_mass(g)
     !enddo
-    !write(*,*)'Liquid mass pre,post,diff ',step,mass_pre,mass_post,mass_pre-mass_post
+    write(*,*)'Liquid mass pre,post,diff ',step,mass_pre,mass_post,mass_pre-mass_post
 
     step_mod = mod(step,48);
     write(string,*) step
@@ -462,6 +493,12 @@ implicit none
      call TDyOutputRegression(tdy,U,ierr);
      CHKERRA(ierr);
   end if
+
+  ! Clean up diagnostic stuff
+  call VecDestroy(diags, ierr)
+  CHKERRA(ierr)
+  call DMDestroy(diags_dm, ierr)
+  CHKERRA(ierr)
 
   call TDyDestroy(tdy,ierr);
   CHKERRA(ierr);
--- a/demo/transient/transient_snes_mpfaof90.F90
+++ b/demo/transient/transient_snes_mpfaof90.F90
@@ -1,6 +1,7 @@
 module snes_mpfaof90mod
 
   use tdycore
+  use petscdm
 #include <petsc/finclude/petsc.h>
 #include <finclude/tdycore.h>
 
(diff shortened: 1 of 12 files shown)


jeff-cohere:
There's one remaining issue with this PR, which I've asked Jed about in #217

demo/transient/transient_snes_mpfaof90.F90
```diff
@@ -376,6 +377,15 @@ program main
   call TDySetInitialCondition(tdy,U,ierr);
   CHKERRA(ierr);
 
+  ! Initialize the diagnostics DM and Vec, fetch the section that holds the
+  ! layout, and determine which diagnostic index corresponds to the liquid
+  ! mass (to check mass conservation).
+  call TDyCreateDiagnostics(tdy, diags_dm, ierr)
```
bishtgautam, line 383: Can `diags_dm` be part of the TDycore lib itself?

demo/transient/transient_snes_mpfaof90.F90
```diff
@@ -428,14 +443,19 @@ program main
     call TDyPostSolveSNESSolver(tdy,U,ierr);
     CHKERRA(ierr);
 
-    ! TODO: Mass conservation diagnostic output is disabled for the moment.
+    ! Check the total liquid mass after the step. Because the liquid mass is
+    ! nonnegative, the sum over all cells is the same as the 1-norm.
+    call TDyComputeDiagnostics(tdy, diags_dm, diags, ierr)
```
bishtgautam, line 448: 1. `TDyComputeDiagnostics` is getting a strided `Vec` with all diagnostics. So, how about renaming it to `TDyGetDiagnostics`?
2. Apart from renaming as suggested above, how about also doing the following?
   - Create another function named `TDyComputeDiagnostics` that fills all the diagnostics in a strided `Vec diags`, where `diags` is owned by TDycore.
   - Then, a function like `TDyGetSaturation` can use `VecStrideGather` to extract the subset from `diags`?

demo/transient/transient_snes_mpfaof90.F90
jeff-cohere: It can. The reason I was doing things this way is to avoid incurring the cost of creating the diagnostics DM in cases where it's not necessary, and also to make clear that the DM can be manipulated by a user without affecting the dycore.

bishtgautam: > avoid incurring the cost of creating the diagnostics DM in cases where it's not necessary
Makes

demo/transient/transient_snes_mpfaof90.F90
jeff-cohere: `TDyComputeDiagnostics` is actually doing what your suggested `TDyComputeDiagnostics` function would do. It doesn't allocate or get the `diags` vector--that is done on line 385, using the DM created by `TDyCreateDiagnostics`. `TDyComputeDiagnostics` simply computes and fills in the strided diagnostics vector.

Regarding `TDyGetSaturation`: Jed says we have to create a "sub-DM" from `diags_dm` that can create a vector that stores only the saturation vector. That would have to be done somewhere in order to get a vector that can hold the saturation (to be passed to `TDyGetSaturation`), so we're looking at another function to create a sub vector (from the sub-DM) for each diagnostic. It seems a little cumbersome to me to have all these DMs floating around, but maybe I'm just not used to manipulating function spaces in the same way as we manipulate fields.

To summarize, we currently have:

1. `TDyCreateDiagnostics`: creates a DM that manages the layout for all diagnostic fields (from which a strided diagnostics vector can be created using `DMCreateGlobalVector`).
2. `TDyComputeDiagnostics`: given the diagnostics DM and a diagnostics vector, populates the vector with the updated diagnostics fields.

I think (in your first comment) you are suggesting that we pull `TDyCreateDiagnostics` inside the dycore itself so that we can eliminate the diagnostics DM from the interface. I think that's fine, except that it then becomes harder to access the layout for the diagnostics fields, so the diagnostics vector becomes a "black box". Tell me what you think of this:

1. `TDyCreateDiagnosticsVector(TDy tdy, Vec *diag_vec)`: creates a global vector `diag_vec` with the proper layout for the dycore's diagnostics fields. It's a black box, as its layout is defined by the diagnostics DM, which isn't exposed to the user.
2.  `TDyCreateDiagnosticsSubVector(TDy tdy, PetscInt index, Vec *sub_vec)`: creates a global vector that stores a single diagnostic field (like liquid mass or saturation) with the given field index within the (hidden) diagnostics DM.
3. `TDyComputeDiagnostics(TDy tdy, Vec diag_vec)`: computes all diagnostic fields, storing the computed values in `diag_vec`. This is the same as what we have now, except we exclude the diagnostics DM from the interface.
4. `TDyGetSaturation(TDy tdy, Vec diag_vec, Vec sat_vec)`: extracts the saturation values from `diag_vec` into `sat_vec` (using e.g. `VecStride`).

bishtgautam: > It seems a little cumbersome to me to have all these DMs floating around

I agree that is why I think the DM and sub-DM should be internal to TDycore. These DMs can be created the first time user calls `TDyComputeDiagnostics`.

Regarding the proposed approach steps 1-5:

- It seems to me if I'm only interested in `TDyGetSaturation`, I have to also do steps 1-3. It is not a big deal for now as we don't have any user base, but it kinda becomes cumbersome.

jeff-cohere: The reason I was proposing  `diag_vec` in addition to the liquid mass and saturation vectors is that we can compute all diagnostic fields at once, which is likely cheaper (especially on GPUs) than recomputing the saturation and liquid mass every time you ask for each of them. You should only have to do step 1 once in your program, and step 2 only once per diagnostic field to be extracted.

If this is confusing or cumbersome, we could consider some rules for how/when diagnostic fields are recomputed, and have the dycore store what we've been calling `diag_vec` internally. For example, we could say that diagnostic fields are updated every time a solver advances by a single time step. Or, we could have a function that explicitly updates all diagnostic fields internally so you can extract the most up-to-date values, e.g.

```
// Update all diagnostics fields given the current state of the dycore.
PetscErrorCode TDyUpdateDiagnostics(TDy tdy);
```

Then we would provide a functions for creating a Vec that can hold cell-centered scalar diagnostic fields (for example):

```
// Create a Vec that can store a cell-centered diagnostic field with a single component.
PetscErrorCode TDyCreateDiagnosticVector(TDy tdy, Vec *vec);
```

If we need vector-valued diagnostics, like the momentum vector within each cell, we could add an argument for the number of components in the diagnostic field. Finally, we can add functions for getting the saturation and liquid masses for each cell:

```
// Get the saturation in each cell, given a Vec created by TDyCreateDiagnosticVector.
PetscErrorCode TDyGetSaturation(TDy tdy, Vec sat_vec);
// Get the liquid mass contained in each cell, given a Vec created by TDyCreateDiagnosticVector.
PetscErrorCode TDyGetLiquidMass(TDy tdy, Vec liq_mass_vec);
```

So we would still have 4 functions, and the user still has to create the `Vec`s that store the saturation and liquid mass. But I don't think we could really simplify things any further than that, because the dycore doesn't know how many `Vec`s a user wants to use to store diagnostics.

How does this sound to you?

bishtgautam: @jeff-cohere I like your proposed approach of:
1. `TDyUpdateDiagnostics() // internal code`
2. `TDyCreateDiagnosticVector() // user code`
3. `TDyGetXYZ() // user code`

jeff-cohere:
Okay, here's the updated version of our diagnostics field calculations. The interface consists of these functions:

* `TDyUpdateDiagnostics(TDy tdy)`: tells the dycore to update the diagnostic fields based on its state
* `TDyCreateDiagnosticVector(TDy tdy, Vec * vec)`: creates a Vec that can store a single cell-centered scalar diagnostic field
* `TDyGetSaturation(TDy tdy, Vec sat)`: populates `sat` with the current liquid saturation values
* `TDyGetLiquidMass(TDy tdy, Vec mass)`: populates `mass` with the current liquid mass values

@bishtgautam , I think you preferred to make `TDyUpdateDiagnostics` an internal function, but this requires us to decide when the diagnostics are updated automatically, and it's not clear to me that we know how to do this for all cases of interest. So can we keep it public for now and then tuck it inside if/when we decide it can be done automatically?

jeff-cohere:
Hey @bishtgautam , have you had a chance to look at this yet? If the interface looks okay to you, we can get it merged.

src/tdycore.c
```diff
@@ -1333,11 +1371,106 @@ PetscErrorCode TDySetDtimeForSNESSolver(TDy tdy, PetscReal dtime) {
...
+}
+
+/// Given a Vec created with TDyCreateDiagnosticVector, populates that vector
+/// with the saturation values for each cell in the grid.
+/// @param [in] tdy A TDy object
+/// @param [out] sat_vec The Vec that stores the cell-centered saturation field
+PetscErrorCode TDyGetSaturation(TDy tdy, Vec sat_vec) {
+  PetscErrorCode ierr;
+  PetscFunctionBegin;
+  ierr = ExtractDiagnosticField(tdy, 0, sat_vec); CHKERRQ(ierr);
```
bishtgautam, line 1462: Regarding the `index` (`= 0`), can we have a variable name that is obvious? Not sure if we can have a global variable e.g. `DIAG_SATURATION`. Another option would be to pass the field name (`"Saturation"`) and determine the offset for that field using `PetscSectionGetFieldOffset`.

bishtgautam:
I only have one minor comment. Let me know if it is something you would like to address in this PR, otherwise, this is ready to be merged.

src/tdycore.c
jeff-cohere: I changed the name to "LiquidSaturation" and `TDyGetLiquidSaturation` to match the liquid mass stuff, and introduced these diagnostic field index macros:

* `DIAG_LIQUID_SATURATION` (0)
* `DIAG_LIQUID_MASS` (1)

Good suggestion.

bishtgautam (approved):

PR #221, opened 2022-01-19
Fixes for parallel runs

bishtgautam:
This PR contains multiple bug fixes for parallel runs.

### DIFF
--- a/src/fe/tdywy.c
+++ b/src/fe/tdywy.c
@@ -288,7 +288,10 @@ PetscErrorCode TDySetFromOptions_WY(void *context, TDyOptions *options) {
   // Set characteristic curve data.
   wy->vangenuchten_m = options->vangenuchten_m;
   wy->vangenuchten_alpha = options->vangenuchten_alpha;
-  wy->mualem_poly_low = options->mualem_poly_low;
+  wy->mualem_poly_x0 = options->mualem_poly_x0;
+  wy->mualem_poly_x1 = options->mualem_poly_x1;
+  wy->mualem_poly_x2 = options->mualem_poly_x2;
+  wy->mualem_poly_dx = options->mualem_poly_dx;
 
   PetscFunctionReturn(0);
 }
@@ -465,20 +468,30 @@ PetscErrorCode TDySetup_WY(void *context, DM dm, EOS *eos,
 
   // By default, we use the the Mualem relative permeability model.
   {
-    PetscReal parameters[6*nc];
+    PetscInt num_params = 9;
+    PetscReal parameters[num_params*nc];
     for (PetscInt c = 0; c < nc; ++c) {
-      PetscReal m = wy->vangenuchten_m,
-                poly_low = wy->mualem_poly_low;
-      parameters[6*c]   = m;
-      parameters[6*c+1] = poly_low;
+      PetscReal m = wy->vangenuchten_m;
+      PetscReal poly_x0 = wy->mualem_poly_x0;
+      PetscReal poly_x1 = wy->mualem_poly_x1;
+      PetscReal poly_x2 = wy->mualem_poly_x2;
+      PetscReal poly_dx = wy->mualem_poly_dx;
+
+      PetscInt offset = num_params*c;
+      parameters[offset    ]   = m;
+      parameters[offset + 1] = poly_x0;
+      parameters[offset + 2] = poly_x1;
+      parameters[offset + 3] = poly_x2;
+      parameters[offset + 4] = poly_dx;
 
       // Set up cubic polynomial coefficients for the cell.
       PetscReal coeffs[4];
-      RelativePermeability_Mualem_GetSmoothingCoeffs(m, poly_low, coeffs);
-      parameters[6*c+2] = coeffs[0];
-      parameters[6*c+3] = coeffs[1];
-      parameters[6*c+4] = coeffs[2];
-      parameters[6*c+5] = coeffs[3];
+      ierr = RelativePermeability_Mualem_GetSmoothingCoeffs(m, poly_x0, poly_x1, poly_x2, poly_dx, coeffs);
+      CHKERRQ(ierr);
+      parameters[offset + 5] = coeffs[0];
+      parameters[offset + 6] = coeffs[1];
+      parameters[offset + 7] = coeffs[2];
+      parameters[offset + 8] = coeffs[3];
     }
     ierr = RelativePermeabilitySetType(cc->rel_perm, REL_PERM_FUNC_MUALEM, nc,
                                        points, parameters); CHKERRQ(ierr);
@@ -1072,13 +1085,13 @@ PetscErrorCode TDyComputeErrorNorms_WY(void *context, DM dm, Conditions *conditi
 
 PetscErrorCode TDyUpdateState_WY(void *context, DM dm,
                                  EOS *eos, MaterialProp *matprop,
-                                 CharacteristicCurves *cc, PetscReal *U) {
+                                 CharacteristicCurves *cc,
+                                 PetscInt num_cells, PetscReal *U) {
   PetscErrorCode ierr;
   PetscFunctionBegin;
   TDyWY *wy = context;
 
-  PetscInt cStart, cEnd;
-  ierr = DMPlexGetHeightStratum(dm,0,&cStart,&cEnd); CHKERRQ(ierr);
+  PetscInt cStart = 0, cEnd = num_cells;
   PetscInt nc = cEnd - cStart;
 
   // Compute the capillary pressure on all cells.
@@ -1127,6 +1140,8 @@ PetscErrorCode TDyWYResidual(TS ts,PetscReal t,Vec U,Vec U_t,Vec R,void *ctx) {
   TDyWY *wy = tdy->context;
 
   ierr = TSGetDM(ts,&dm); CHKERRQ(ierr);
+  ierr = DMPlexGetHeightStratum(dm,0,&cStart,&cEnd); CHKERRQ(ierr);
+  ierr = DMPlexGetHeightStratum(dm,1,&fStart,&fEnd); CHKERRQ(ierr);
   nv   = wy->nfv;
   wgt  = 1/((PetscReal)nv);
   ierr = DMGetLocalVector(dm,&Ul); CHKERRQ(ierr);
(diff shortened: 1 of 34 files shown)


jeff-cohere:
Thanks, @bishtgautam . Probably a lot of these issues were oversights on my part when I was refactoring.

One question: I see you've added an `ncells` parameter to `TDyUpdateState` and its related functions. Because the first argument is a DM, it's possible to get the number of cells without this extra argument. And if `ncells` disagrees with the DM about the number of cells, we will get unexpected behavior. So I would recommend not having this extra parameter, unless I'm missing something.

src/fe/tdywy.c
```diff
@@ -1072,13 +1072,13 @@ PetscErrorCode TDyComputeErrorNorms_WY(void *context, DM dm, Conditions *conditi
...
 PetscErrorCode TDyUpdateState_WY(void *context, DM dm,
                                  EOS *eos, MaterialProp *matprop,
-                                 CharacteristicCurves *cc, PetscReal *U) {
+                                 CharacteristicCurves *cc,
+                                 PetscInt num_cells, PetscReal *U) {
   PetscErrorCode ierr;
   PetscFunctionBegin;
   TDyWY *wy = context;
 
-  PetscInt cStart, cEnd;
```
jeff-cohere, line 1080: Are we getting the wrong number of cells here? Why do we have to pass the number of cells to this function?

src/mpfao/tdympfao_snes.c
```diff
@@ -43,7 +43,7 @@ PetscErrorCode TDyMPFAOSNESPreSolve(TDy tdy) {
 
   // Update the auxillary variables
   ierr = VecGetArray(tdy->soln_prev,&p); CHKERRQ(ierr);
-  ierr = TDyUpdateState(tdy, p); CHKERRQ(ierr);
+  ierr = TDyUpdateState(tdy, p, mesh->num_cells_local); CHKERRQ(ierr);
   ierr = VecRestoreArray(tdy->soln_prev,&p); CHKERRQ(ierr);
```
bishtgautam, line 47: @jeff-cohere You have a good point regarding avoiding the need of passing the third argument i.e.`mesh->num_cells_local`. I will change it by passing `TDyUpdateState` the `Vec` on which the function needs to perform computation.

bishtgautam: The proposed refactor is causing WY to fail. The refactored branch is `bishtgautam/fix-parallel-runs-rebase-02-02-2022-updatestate`. I haven't been able to debug the error. Can we merge this PR and open a new issue to track this review comment?

jeff-cohere:
I created #224. Let me do a review of this PR so we can get it merged.

src/materials/tdycharacteristiccurves.c
```diff
@@ -504,20 +505,46 @@ void RelativePermeability_Mualem_Unsmoothed(PetscReal m,PetscReal Se,PetscReal *
...
                                            PetscReal rhs[4]) {
 
-  PetscInt n = 4, nrhs = 1;
+  PetscInt N = 4, d;
+  PetscInt *indx;
+  PetscReal **AA;
+  PetscReal *rhs_values;
+  PetscErrorCode ierr;
+
+  ierr = TDyAllocate_RealArray_2D(&AA,N,N); CHKERRQ(ierr);
```
jeff-cohere, line 514: These need to be deallocated at the end of the function to avoid a memory leak. There are other ways to allocate "work arrays" in C (even multidimensional ones). Maybe we should chat about that when it's convenient.

src/mpfao/tdymesh.c
```diff
@@ -963,16 +948,20 @@ static PetscErrorCode IdentifyLocalCells(DM dm, TDyMesh *mesh) {
...
     if (gref>=0) {
       cells->is_local[c] = PETSC_TRUE;
       cells->global_id[c] = gref;
+      num_cells_local++;
     } else {
       cells->is_local[c] = PETSC_FALSE;
       cells->global_id[c] = -gref-1;
     }
   }
+  mesh->num_cells_local = num_cells_local;
```
jeff-cohere, line 964: According to this definition of "local cell", `mesh->num_cells_local != (cEnd - cStart)`. This is probably an important clue to what's going on in #224.

src/mpfao/tdympfao.c
```diff
@@ -1950,6 +1950,96 @@ static PetscErrorCode ComputeFacePermeabilityTensor(TDyMPFAO *mpfao,
...
+///   along the line joining up and down cells
+/// - Third, dot producot of the projected face permeability vector and unit
+///   normal to the face is computed
+///
+/// @param [in] tdy A TDy struct
+/// @param [in] dim dimension of the problem
+/// @param [in] face_id ID of the face
+/// @param [out] *k_face_value Permeability value at the face
+/// @returns 0 on success, or a non-zero error code on failure
+PetscErrorCode ComputeFacePeremabilityValueMPFAO(TDyMPFAO *mpfao, MaterialProp *matprop, PetscInt dim, PetscInt face_id, PetscReal *Kface_value) {
```
jeff-cohere, line 2011: Peremability -> Permeability

src/tdyutils.c
```diff
@@ -482,6 +482,31 @@ PetscInt TDyReturnIndexInList(PetscInt *list, PetscInt nlist, PetscInt value) {
 }
 
 /* -------------------------------------------------------------------------- */
+/// Outputs a sequential PETSc Vec as binary
+///
+/// @param [in] vec A sequential PETSc Vec
+/// @param [in] filename Name of the file
+/// @returns 0 on success, or a non-zero error code on failure
+PetscErrorCode TDySavePetscVecSeqAsBinary(Vec vec, const char filename[]) {
```
jeff-cohere, line 490: I don't think we need this function. Instead, we could modify `TDySavePetscVecAsBinary` to get the communicator for the given Vec and use it instead of `PETSC_COMM_WORLD` there. Then it would work regardless of the type of Vec.

Let me know if you'd like me to do this.

src/tdyutils.c
```diff
@@ -616,3 +641,128 @@ PetscErrorCode ComputeInverseOf3by3Matrix(PetscReal a[9], PetscReal inv_a[9]) {
   PetscFunctionReturn(0);
 }
 
+/* -------------------------------------------------------------------------- */
+/// Performs a LU decomposition of a matrix
+///
+/// The function is from "Numerical Recipes in C"
```
jeff-cohere, line 647: Numerical Recipes has horrible licensing conditions, so I don't think we want to paste their stuff into TDycore. I think we should rely on LAPACK here instead. I'm sure we're building it as part of PETSc.

jeff-cohere (changes requested):
Most of this looks good, but I noticed a few issues that are worth addressing before we merge this. I'm happy to handle some or all of these.

src/tdyutils.c
bishtgautam: I agree it should be deleted.

src/tdyutils.c
bishtgautam: Is it still a violation of license if the code is not a copy/paste but written looking at the code provided in the book?

jeff-cohere: Yes: http://numerical.recipes/licenses/redistribute.html

bishtgautam: There was a reason why I implemented the code from NR instead of using LAPACK, but I can't recall it. It might be that I needed LU of NR to match PFLOTRAN's implementation. I will test out LAPACK.

jedbrown: How big are the matrices here? PETSc has some unrolled block kernels and of course you can use LAPACK. I agree that NA is problematic, both on licensing and reliability.
https://www.stat.uchicago.edu/~lekheng/courses/302/wnnr/nr.html

bishtgautam: These are 4x4 matrices. I was using NA because of the discussion that started the following email thread https://groups.google.com/g/tdycores-dev/c/TXdbtr7Oo_I/m/zgDWAdFmCAAJ. I'm now working on your suggested solution of using transformation on that email thread.

bishtgautam: @jedbrown's suggested [variable transformation](https://github.com/TDycores-Project/TDycore/pull/221/files#diff-af758c4df5c826801009d3edcf5374849f379423e4115db5a18dfc8445f9a83cR504) significantly improves the condition number of the matrix and testing shows that cubic polynomial coeffs are computed such that `kr<=1`.

bishtgautam:
To remove the NR code, I squashed all commits into a single commit.

src/materials/tdycharacteristiccurves.c
```diff
@@ -531,16 +546,18 @@ static PetscErrorCode CubicPolynomialSetup(PetscReal x1, PetscReal x2,
...
                                               PetscReal *f, PetscReal *df_dx) {
 
   PetscFunctionBegin;
 
-  *f = coeffs[0] + coeffs[1]*x + coeffs[2]*PetscPowReal(x,2.0) + coeffs[3]*PetscPowReal(x,3.0);
-  *df_dx = coeffs[1] + 2.0*coeffs[2]*x + 3.0*coeffs[3]*PetscPowReal(x,2.0);
+  PetscReal xt = (x-x0)/dx;
+
+  *f = coeffs[0] + coeffs[1]*xt + coeffs[2]*PetscPowReal(xt,2.0) + coeffs[3]*PetscPowReal(xt,3.0);
+  *df_dx = coeffs[1]/dx + 2.0*coeffs[2]*xt/dx + 3.0*coeffs[3]*PetscPowReal(xt,2.0)/dx;
```
jedbrown, line 560: A bit more efficient to use Horner's rule
```suggestion
  *f = coeffs[0] + xt * (coeffs[1] + xt * (coeffs[2] + xt * coeffs[3]));
  *df_dx = (coeffs[1] + xt * (2.0*coeffs[2] + xt * 3.0 * coeffs[3])) / dx;
```

jedbrown (approved):

src/mpfao/tdympfao.c
jeff-cohere: @bishtgautam , can you fix this typo? It's not a big deal, but it's easier to fix now than later.

src/mpfao/tdympfao.c
```diff
@@ -2041,31 +2131,31 @@ static PetscErrorCode ComputeGravityDiscretization(TDyMPFAO *mpfao, DM dm,
...
-      // Ku = K_face x u_up2dn
-      PetscReal Ku[dim];
-      for (PetscInt ii = 0; ii < dim; ii++ ){
-        Ku[ii] = 0.0;
-        for (PetscInt jj = 0; jj < dim; jj++ ){
-          Ku[ii] += K_face[ii*dim + jj] * u_up2dn[jj];
-        }
-      }
+      case MPFAO_GMATRIX_TPF:
+        ierr = ComputeFacePeremabilityValueTPF(mpfao, matprop, dim, face_id, &k_face_value); CHKERRQ(ierr);
```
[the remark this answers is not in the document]
jeff-cohere: This should be updated when you fix the function name above.

PR #226, opened 2022-04-04
Adds two-point flux as a new discretization

bishtgautam:
Adds two-point flux as a separate discretization following the approach used
in PFLOTRAN.


bishtgautam:
@jeff-cohere This is ready for your review.

gmakefile
```diff
@@ -12,7 +12,7 @@ libtdycore_static := $(LIBDIR)/libtdycore.$(AR_LIB_SUFFIX)
 libtdycore := $(if $(filter-out no,$(BUILDSHAREDLIB)),$(libtdycore_shared),$(libtdycore_static))
 
 pkgs := tdycore
-srcs-tdycore.c   := $(wildcard src/mpfao/*.c src/fe/*.c src/materials/*.c src/*.c src/f90-mod/*.c)
```
jeff-cohere, line 15: So does this mean we're just not going to compile the finite element code for now?

include/private/tdyconditionsimpl.h
```diff
@@ -22,6 +23,9 @@ typedef struct Conditions {
   /// Compute the pressure at a set of given points on the boundary.
   PetscErrorCode (*compute_boundary_pressure)(void*,PetscInt,PetscReal*,PetscReal*);
 
+  /// Set the pressure at a set of given points on the boundary.
+  PetscErrorCode (*assign_boundary_pressure_type)(void*,PetscInt,PetscReal*,PetscInt*);
```
jeff-cohere, line 27: I guess I still don't quite understand why we need this. Does the two-point flux approximation actually support Neumann pressure boundary conditions in addition to the Dirichlet/seepage types? Wouldn't a "Neumann" pressure boundary condition be related linearly to a boundary condition on the normal velocity? If so, we could check the `compute_boundary_pressure` and `compute_boundary_velocity` function pointers to determine the type of boundary condition. Specifically:

* If `compute_boundary_pressure` and `compute_boundary_velocity` are both `NULL`, we have a seepage BC.
* If `compute_boundary_pressure` is non-`NULL`, we have a Dirichlet pressure BC.
* If `compute_boundary_velocity` is non-`NULL`, we have a Neumann pressure BC.

Might that be simpler than adding a different function type? If needed we could add a helper function to check these pointers and return something that indicates the "pressure BC type". But I very well could be missing something here.

gmakefile
bishtgautam: We are still building the FE code. I added a new dir `src/fvtpf/*.c`

jeff-cohere: Ah, right. I see now.

include/private/tdyfvtpfimpl.h
```diff
@@ -0,0 +1,103 @@
...
+  PetscReal  *drho_dT, *dvis_dT;
+
+  /* problem constants */
+  PetscReal  gravity[3]; /* vector of gravity [m s-2] */
+  PetscReal  Pref;       /* reference pressure */
+  PetscReal  Tref;       /* reference temperature */
+
+} TDyFVTPF;
+
+// Functions specific to MPFA-O implementations.
```
jeff-cohere, line 81: specific to TPF implementations, really, no?

include/tdycore.h
```diff
@@ -43,12 +45,12 @@ typedef enum {
 PETSC_EXTERN const char *const TDyMPFAOGmatrixMethods[];
 
 typedef enum {
-  MPFAO_DIRICHLET_BC=0,  /* Dirichlet boundary condiiton */
-  MPFAO_NEUMANN_BC,       /* Neumann zero-flux boundary condition */
-  MPFAO_SEEPAGE_BC       /* Seepage boundary condition */
-} TDyMPFAOBoundaryConditionType;
+  DIRICHLET_BC=0,  /* Dirichlet boundary condiiton */
```
jeff-cohere, line 48: This is fine, but we're probably going to replace it with the new boundary condition logic.

jeff-cohere:
It feels like a lot of this code is going to be very, very temporary, since we're changing the way that we specify, track, and enforce boundary conditions. On the other hand, it's nice to have a proper two-point flux approximation.

The biggest question I have is whether it's worth adding all of the machinery for supporting integer-valued functions if we can get by without doing it (see my longest comment above). As far as I can tell, we can already determine the "BC type" for the entire domain using specified pressure/velocity/neither boundary conditions (or at least we could do so with modifications to make it clear that when you don't specify P or Vn, you're asking for a seepage condition).

What do you think, @bishtgautam ? Do you want to just add the integer-valued "BC type" functions for now? They'll go away as soon as the new BC code comes in, but if things are working now, maybe it's easiest to keep what's in this PR. Your call.

bishtgautam:
I would like to keep the integer-value BC for now as I want to see how things will evolve when we have all BC-related info being read from a `.exo` file.

include/private/tdyconditionsimpl.h
bishtgautam: - If no BC is specified on a face, it will be a Neumann with zero-flux. (Currently, the Neumann BC implies a zero-flux. Eventually, an non zero-flux Neumann BC needs to be supported.)
- This is indeed a temporary way of adding dirichlet and seepage BC in the domain. 
- After your changes are merged, I envision there would be support for a user-defined function that adds a BC condition: (1) BC-type and (2) values of the BC on a collection of faces.

demo/transient/fv_tpff90.F90
```diff
@@ -0,0 +1,334 @@
...
+  ! Set a constructor for a DM.
+  call TDySetDMConstructor(tdy, CreateDM,ierr); CHKERRA(ierr)
+
+  ! Apply overrides.
+  call TDySetFromOptions(tdy,ierr); CHKERRA(ierr)
+
+  call TDyGetDM(tdy, dm, ierr); CHKERRA(ierr)
+  call DMGetDimension(dm, dim, ierr);  CHKERRA(ierr)
+  call PetscFECreateDefault(PETSC_COMM_SELF, dim, 1, PETSC_FALSE, "p_", -1, fe, ierr); CHKERRA(ierr)
+  call PetscObjectSetName(fe, "p", ierr); CHKERRA(ierr)
```
jeff-cohere, line 253: Is this `p` field the MPI rank of the process? Does our model assume this field exists? If so, maybe we could fold this code into `TDySetup`.

src/fvtpf/tdyfvtpf.c
```diff
@@ -0,0 +1,1029 @@
...
+                                  &(fvtpf->d2rho_dP2[c])); CHKERRQ(ierr);
+    ierr = EOSComputeWaterViscosity(eos, P, &(fvtpf->vis[c]),
+                                    &(fvtpf->dvis_dP[c]),
+                                    &(fvtpf->d2vis_dP2[c])); CHKERRQ(ierr);
+  }
+
+  PetscFunctionReturn(0);
+}
+
+static PetscErrorCode ComputeGeometry(TDyFVTPF *fvtpf, DM dm) {
```
jeff-cohere, line 516: How does this function relate to the function you added to `TDyMesh` that computes geometry?

src/fvtpf/tdyfvtpf_snes.c
```diff
@@ -0,0 +1,714 @@
...
+    if (faces->is_internal[iface]) continue; // skip internal faces
+    if (faces->bc_type[iface] == NEUMANN_BC) continue; // skip non-flux faces
+
+    PetscInt *cell_ids, num_face_cells;
+    ierr = TDyMeshGetFaceCells(mesh, iface, &cell_ids, &num_face_cells); CHKERRQ(ierr);
+
+    PetscInt cell_id_dn = cell_ids[1];
+    PetscReal Jdn;
+    switch (faces->bc_type[iface]) {
+    case NEUMANN_BC:
```
jeff-cohere, line 676: This case is excluded above. Can you change this to a `default` block and put it below the other cases?

src/fvtpf/tdyfvtpf_snes.c
```diff
@@ -0,0 +1,714 @@
...
+    ierr = TDyMeshGetFaceCells(mesh, iface, &cell_ids, &num_face_cells); CHKERRQ(ierr);
+    PetscInt cell_id_up = cell_ids[0];
+    PetscInt cell_id_dn = cell_ids[1];
+
+    PetscReal Jup, Jdn;
+    ierr = RichardsJacobian(fvtpf, tdy->dm, tdy->matprop, iface, &Jup, &Jdn);
+
+    if (cell_id_up >= 0 && cells->is_local[cell_id_up]) {
+      //r_ptr[cell_id_up] += Res;
+      ierr = MatSetValuesLocal(B,1,&cell_id_up,1,&cell_id_up,&Jup,ADD_VALUES);CHKERRQ(ierr);
```
jeff-cohere, line 650: These could be vectorized to improve performance. Not critical for this PR, though.

jeff-cohere (approved):
This stuff looks okay. I left a few questions and comments. I think there's still too much copying/pasting to add a new discretization, but we are working our way toward making this easier.

bishtgautam:
I have now deleted the `ComputeGeometry` from MPFA-O and FV-TPF and replaced with `TDyMeshComputeGeometry`.  (Though my head hurts because of all `&` `*` `[]` in [the function](https://github.com/TDycores-Project/TDycore/pull/226/commits/45cc2253504e83b066bf85ebc7f1ea44a28749d4#diff-3dd056d541e4f2df898c815798be188f185c120cee8af7d4b714f710908c4ee7R4862))

PR #230, opened 2022-05-27
[WIP] Adds a DM to manage unstructured grid using PFLOTRAN's approach

bishtgautam:


include/private/tdydiscretizationimpl.h
```diff
@@ -0,0 +1,32 @@
...
+#define TDYDISCRETIZATION_H
+
+#include <petsc.h>
+#include <private/tdydmimpl.h>
+
+
+typedef struct {
+
+  TDyDM *tdydm;
+  TDyUGrid *ugrid;
```
jeff-cohere, line 11: I remember you talking a bit about the DM "shell" type, and that you were having some issues using it. Can you remind me of the issues you encountered? It seems like it would be more convenient to create a DM shell for the PFlotran-style  FV mesh and use the "ugrid" type as its "context". Then we wouldn't have to carry around this additional `TDyDiscretization` type--we could just check whether the DM is a shell, and if it is, we know it's a PFlotran mesh, and we can extract the context to get the ugrid data.

src/f90-mod/tdycoref.c
```diff
@@ -24,6 +24,7 @@
 #define tdyupdatediagnostics_                          TDYUPDATEDIAGNOSTICS
 #define tdycreatediagnosticvector_                     TDYCREATEDIAGNOSTICVECTOR
 #define tdycreateprognosticvector_                     TDYCREATEPROGGNOSTICVECTOR
+#define tdycreatepmatrix_                              TDYCREATEMATRIX
```
jeff-cohere, line 27: Should `tdycreatepmatrix` be `tdycreatematrix` here?

makefile
```diff
@@ -7,6 +7,10 @@ ifdef codecov
   LIBS    += -lgcov
 endif
 
+ifdef ugrid_debug
```
bishtgautam, line 10: @jeff-cohere With this change, I wanted to allow for the addition of a compiler flag `-DUGRID_DEBUG`, but here is the issue:

- `make ugrid=1 V=1`: The new compiler flag isn't used
- `make ugrid=1 V=1 all-gmake`: The new compiler flag is used

Any what is going wrong?

jeff-cohere: I'm not sure. It might be related to the fact that the first command invokes the second indirectly, but maybe doesn't capture the `ugrid=1` flag you're passing. If that's true,  we need to explicitly pass along arguments from the default (`all`) target to `all-gmake`.

More broadly: this build system has worked for us for a while now, but it's still pretty confusing (we have all of the different spellings of makefiles, for example), and it's not clear to a newcomer what gets invoked where unless said newcomer is versed in all of Make's various assumptions (which fewer and fewer people are as time passes).

As we move toward becoming part of a larger ecosystem, we might need a more "turnkey" build system that is a little easier for a non-expert to extend.

jeff-cohere: Yeah, I'm not complaining! This build system has taken the project pretty far, and I appreciate not having had to touch it very much. Build systems are difficult in general, and it's always difficult to strike a balance between dealing with existing quirks and spending time improving the system itself.

jedbrown: Note that gmake is quite a bit simpler if you always build from the top level because it sees all the dependencies. It's clutter and harder to test if you want stand-alone makefiles in each subdirectory. Sometimes people want subdirectories to be stand-alone because the idea is that a user would copy them somewhere else as step one of a new project. If we need to refactor/change the build, it would be useful to gather requirements.

demo/richards/richards_driver.c
```diff
@@ -31,9 +31,7 @@ int main(int argc, char **argv) {
   ierr = TDySetFromOptions(tdy); CHKERRQ(ierr);
 
   ierr = TDyDriverInitializeTDy(tdy); CHKERRQ(ierr);
-  if (!rank) {
-    ierr = TDyIOSetIOProcess(tdy->io, PETSC_TRUE); CHKERRQ(ierr);
-  }
+  ierr = TDyIOSetIOProcess(tdy->io, PETSC_TRUE); CHKERRQ(ierr);
```
jeff-cohere, line 34: Is there a reason you want I/O on every process instead of just rank 0?

include/private/tdycoreimpl.h
```diff
@@ -55,10 +57,12 @@ struct _TDyOps {
   // before the DM is distributed across processes.
   PetscErrorCode (*set_dm_fields)(void*, DM);
 
+  PetscInt (*get_num_dm_fields)(void*);
```
jeff-cohere, line 60: Can you add a comment above here? It's just easier to have everything documented up front.

include/private/tdydiscretizationimpl.h
jeff-cohere: We'll circle back to this question later.

include/private/tdydiscretizationimpl.h
```diff
@@ -0,0 +1,32 @@
+#if !defined(TDYDISCRETIZATION_H)
+#define TDYDISCRETIZATION_H
+
+#include <petsc.h>
+#include <private/tdydmimpl.h>
+
+
```
jeff-cohere, line 7: @bishtgautam , would you mind adding a brief comment here describing how we're following PFLOTRAN's lead here, and that we may elect to use DMShell down the line?

include/private/tdydmimpl.h
```diff
@@ -0,0 +1,25 @@
+#if !defined(TDYDMIMPL_H)
+#define TDYDMIMPL_H
+
+#include <petsc.h>
+#include <private/tdyugdmimpl.h>
+#include <private/tdyugridimpl.h>
+
+typedef enum{
```
jeff-cohere, line 8: How does this type relate to `TDyDiscretization`? It seems like the same struct with an additional `DM` field.

include/private/tdymeshimpl.h
```diff
@@ -200,9 +213,15 @@ typedef struct TDyMesh {
   TDyRegion region_connected;
 
   PetscInt *closureSize, **closure, maxClosureSize;
+
+  PetscInt *nG2L; // Mapping of global cells to local cells
```
jeff-cohere, line 217: I can't help but wonder if we should make use of PETSc's `IS`-related [types and functions](https://petsc.org/release/docs/manualpages/IS/index.html) for these mappings. This might help us anticipate and avoid issues with global-to-local mappings in large problems.,

include/private/tdyugdmimpl.h
```diff
@@ -0,0 +1,41 @@
...
+  IS IS_GhostCells_in_PetscOrder; // IS of ghost cells from a global Vec
+
+  IS IS_LocalCells_to_NaturalCells; // IS of local cell from a natural Vec
+
+  VecScatter Scatter_LocalCells_to_GlobalCells;   // for scattering data from a local Vec to a global Vec
+  VecScatter Scatter_GlobalCells_to_LocalCells;   // for scattering data from a global Vec to a local Vec
+  VecScatter Scatter_LocalCells_to_LocalCells;    // for scattering data from a local Vec to a local Vec
+  VecScatter Scatter_GlobalCells_to_NaturalCells; // for scattering data from a global Vec to a natural Vec
+
+  ISLocalToGlobalMapping Mapping_LocalCells_to_GhostedCells;
```
jeff-cohere, line 29: This is the mapping type I was referring to in my comment about local <-> global mappings above.

src/fv/fvtpf/tdyfvtpf.c
```diff
@@ -921,8 +765,11 @@ PetscErrorCode FVTPFComputeFacePeremabilityValueTPF(TDyFVTPF *fvtpf, MaterialPro
   PetscErrorCode ierr;
```
jeff-cohere, line 765: Maybe not important, but this function name has a typo in it. Should be `Permeabililty` instead of `Peremability`.

src/tdyti.c
```diff
@@ -104,51 +104,65 @@ PetscErrorCode TDyTimeIntegratorRunToTime(TDy tdy,PetscReal sync_time) {
...
-        ierr = SNESGetLinearSolveIterations(ti->snes,&lit); CHKERRQ(ierr);
-        ierr = SNESGetConvergedReason(ti->snes,&reason); CHKERRQ(ierr);
-        if (reason < 0) {
-          SETERRQ(PETSC_COMM_WORLD,PETSC_ERR_USER,"TDyTimeIntegratorRunToTime failed to converge");
-        }
 
-        if (tdy->io->io_process) {
-          printf("\nTime step %d: time = %f dt = %f ni = %d li = %d rsn = %s\n\n",
+          PetscChar message[1000];
+          sprintf(message,"\nTime step %d: time = %f dt = %f ni = %d li = %d rsn = %s\n\n",
```
jeff-cohere, line 156: As a matter of practice, I recommend using `snprintf` here, which lets you specify the maximum length of the formatted output string (which is 1000 in this case).

src/tdyti.c
```diff
@@ -104,51 +104,65 @@ PetscErrorCode TDyTimeIntegratorRunToTime(TDy tdy,PetscReal sync_time) {
...
             ti->istep,ti->time,ti->dt,nit,lit, SNESConvergedReasons[reason]);
+          ierr = PetscPrintf(PETSC_COMM_WORLD,message); CHKERRQ(ierr);
```
jeff-cohere, line 158: Actually, [PetscPrintf](https://petsc.org/release/docs/manualpages/Sys/PetscPrintf.html) allows you to specify formatting parameters, just like `printf` does. Why not just replace your previous call to `printf` with `PetscPrintf`?

src/tdymemory.c
```diff
@@ -178,6 +178,22 @@ PetscErrorCode TDyAllocate_RealArray_4D(PetscReal *****array_4D, PetscInt ndim_1
   PetscFunctionReturn(0);
 }
 
+/* ---------------------------------------------------------------- */
+PetscErrorCode TDyDeallocate_RealArray_1D(PetscReal *array_1D) {
+
+  PetscFunctionBegin;
+  free(array_1D);
```
jeff-cohere, line 185: We seem to be using `malloc` and `free` in some places and `PetscMalloc` and `PetscFree` in others. Should we decide on one or the other? If so, I'm happy to switch us over to our choice (in this PR or in another).

src/tdycore.c
```diff
@@ -464,11 +478,13 @@ PetscErrorCode TDyView(TDy tdy,PetscViewer viewer) {
...
-/// user. TDySetFromOptions must be called before TDySetup,
-/// since the latter uses options specified by the former.
-/// @param [inout] tdy The dycore instance
-PetscErrorCode TDySetFromOptions(TDy tdy) {
+/// Reads command line options specified by the user
+///
+/// @param [inout] tdy A TDy struct
+///
+/// @returns 0 on success, or a non-zero error code on failure
+PetscErrorCode ReadCommandLineOptions(TDy tdy) {
```
jeff-cohere, line 486: If you want to make this an "internal" function visible only within this file, you can add the `static` keyword before the function's declaration.

src/tdycore.c
```diff
@@ -1833,14 +1950,58 @@ PetscErrorCode TDyPreSolveSNESSolver(TDy tdy) {
   PetscFunctionReturn(0);
 }
 
+/// Resets solver when a time step is cut
+/// @param [inout] TDy struct
+///
+/// @returns 0 on success, or a non-zero error code on failure
+PetscErrorCode TDyTimeCut(TDy tdy) {
```
jeff-cohere, line 1957: Seems like this behavior should be added to our set of "polymorphic" functions.

src/fv/share/tdydm.c
```diff
@@ -0,0 +1,69 @@
...
+///
+/// @returns 0 on success, or a non-zero error code on failure
+PetscErrorCode TDyDMDestroy(TDyDM *tdydm) {
+
+  PetscErrorCode ierr;
+
+  ierr = DMDestroy(&tdydm->dm); CHKERRQ(ierr);
+  ierr = TDyUGDMDestroy(tdydm->ugdm); CHKERRQ(ierr);
+
+  free (tdydm);
```
jeff-cohere, line 31: This should be `PetscFree`, since it was allocated with `PetscCalloc`.

src/fv/share/tdyugrid.c
```diff
@@ -0,0 +1,1209 @@
...
+/// Allocates memory and initializes a TDyUGrid struct
+///
+/// @param [out] ugrid A TDyUGrid struct
+///
+/// @returns 0 on success, or a non-zero error code on failure
+PetscErrorCode TDyUGridCreate(TDyUGrid **ugrid) {
+
+  PetscErrorCode ierr;
+
+  ierr = PetscCalloc(sizeof(TDyUGrid), ugrid); CHKERRQ(ierr);
```
jeff-cohere, line 21: The difference between `malloc` and `calloc` (and `PetscMalloc` and `PetscCalloc`) is that `calloc` zeroes out all memory that is allocated, which makes lines 23-56 unnecessary. (`malloc`, by contrast, just hands you back a chunk of uninitialized memory.)

src/fv/share/tdyugrid.c
```diff
@@ -0,0 +1,1209 @@
...
+  PetscInt *num_verts_local = &ugrid->num_verts_local;
+
+  ierr = PetscViewerHDF5Open(PETSC_COMM_WORLD, mesh_file, FILE_MODE_READ, &viewer);CHKERRQ(ierr);
+
+  ierr = PetscViewerHDF5PushGroup(viewer, "/Domain");CHKERRQ(ierr);
+
+  // Read cells
+  ierr = VecCreate(PETSC_COMM_WORLD, &cells); CHKERRQ(ierr);
+  ierr = PetscObjectSetName((PetscObject) cells, "cells");CHKERRQ(ierr);
+  ierr = PetscObjectSetName((PetscObject) cells, "Cells"); CHKERRQ(ierr);
```
jeff-cohere, line 87: I don't think you can give two names to any single `PetscObject`, so line 87 is likely overwriting the effects of line 86 here. I'd just pick one name and go with it.

src/fv/share/tdyugrid.c
```diff
@@ -0,0 +1,1209 @@
...
+#ifdef UGRID_DEBUG
+  ierr = TDySavePetscISAsASCII(*NewCellRankIS,"is.out");
+#endif
+
+  PetscFunctionReturn(0);
+}
+
+/// Determine the max number of duals/neighbors of a grid cell in the mesh
+///
+/// @param [in] DualMat A dual matrix of the adjaceny matrix
```
jeff-cohere, line 273: `adjaceny` -> `adjacency`

src/fv/share/tdyugrid.c
```diff
@@ -0,0 +1,1209 @@
...
+  *NewNumCellsLocal = cell_counts[myrank];
+
+#ifdef UGRID_DEBUG
+  ierr = TDySavePetscISAsASCII(*NewCellRankIS,"is.out");
+#endif
+
+  PetscFunctionReturn(0);
+}
+
+/// Determine the max number of duals/neighbors of a grid cell in the mesh
```
jeff-cohere, line 271: This determines the max number of duals/neighbors for all cells in the mesh, not just one.

src/fv/share/tdyugrid.c
```diff
@@ -0,0 +1,1209 @@
...
+  PetscInt tmp = ugrid->max_ndual_per_cell;
+  ierr = MPI_Allreduce(&tmp, &ugrid->max_ndual_per_cell, 1, MPI_INTEGER, MPI_MAX, PETSC_COMM_WORLD);
+
+  ierr = MatRestoreRowIJ(DualMat, 0, PETSC_FALSE, PETSC_FALSE, &num_rows, &ia_ptr, &ja_ptr, &success);CHKERRQ(ierr);
+
+  PetscFunctionReturn(0);
+
+}
+
+/// Create IS the maps a natural-ordered Vec to a PETSC-ordered Vec
```
jeff-cohere, line 311: Is "PETSc order" the same as a "global ordering"? If so, let's use that terminology.

src/fv/share/tdyugrid.c
```diff
@@ -0,0 +1,1209 @@
...
+        ugrid->cell_num_vertices[icell] = count;
+        v_ptr[icell*stride + vertex_offset + ivertex] = IntArray4[vertID-1]+1;
+      }
+    }
+  }
+
+  ierr = VecRestoreArray(*LocalOrderVec, &v_ptr); CHKERRQ(ierr); // Now vertex ids are in local-order
+
+#ifdef UGRID_DEBUG
+  char filename[30];
```
jeff-cohere, line 1031: I recommend using something bigger than 30 here, and using `snprintf` instead of `sprintf` to make sure you don't overflow your string.

src/fv/share/tdyugrid.c
```diff
@@ -0,0 +1,1209 @@
...
+  sprintf(filename,"elements_vert_local%d.out",rank);
+  ierr = TDySavePetscVecAsASCII(*LocalOrderVec, filename); CHKERRQ(ierr);
+#endif
+
+  PetscFunctionReturn(0);
+}
+
+/// Save coorindates of local vertices (post-partition) from natural vertices (pre-partition)
```
jeff-cohere, line 1039: coorinates -> coordinates

src/fv/share/tdyugdm.c
```diff
@@ -0,0 +1,477 @@
...
+/// Allocates memory and initializes a TDyUGDM struct
+///
+/// @param [out] ugdm A TDyUGDM struct
+///
+/// @returns 0 on success, or a non-zero error code on failure
+PetscErrorCode TDyUGDMCreate(TDyUGDM **ugdm){
+  PetscFunctionBegin;
+  PetscErrorCode ierr;
+
+  ierr = PetscCalloc(sizeof(TDyUGDM), ugdm); CHKERRQ(ierr);
```
jeff-cohere, line 17: `calloc` and friends zero out the initialized memory, so we don't need lines 19-35.

src/fv/share/tdyugdm.c
```diff
@@ -0,0 +1,477 @@
...
+
+  if (ugdm->Scatter_GlobalCells_to_NaturalCells != NULL) {
+    ierr = VecScatterDestroy(&ugdm->Scatter_GlobalCells_to_NaturalCells); CHKERRQ(ierr);
+  }
+
+  if (ugdm->Mapping_LocalCells_to_GhostedCells != NULL) {
+    ISLocalToGlobalMappingDestroy(&ugdm->Mapping_LocalCells_to_GhostedCells);
+  }
+
+  free(ugdm);
```
jeff-cohere, line 98: Use `PetscFree` here, since you allocated this with `PetscCalloc`.

src/fv/share/tdymeshcustom.c
```diff
@@ -0,0 +1,1315 @@
...
+
+        if (vertex_id > -1) {
+          ugrid->face_to_vertex_natural[ivertex][face_count] = ugrid->vertex_ids_natural[vertex_id];
+        }
+      }
+
+      face_count++;
+    }
+  }
+  free(vertex_ids);
```
jeff-cohere, line 625: This is another example of how we need to unify our memory allocation code (it's not clear in this context that `TDyAllocate_IntegerArray_1D` above uses `malloc` to allocate the memory being freed here).

src/fv/share/tdymeshcustom.c
```diff
@@ -0,0 +1,1315 @@
...
+/// Computes area of triangle
+///
+/// @param [in] ugrid A TDyUGrid struct
+/// @param [in] v_id1 First vertex ID of the triangel
+/// @param [in] v_id2 Second vertex ID of the triangel
+/// @param [in] v_id3 Third vertex ID of the triangel
+/// @param [out] area Area of the triange
+///
+/// @returns 0 on success, or a non-zero error code on failure
+static PetscErrorCode ComputeArea(TDyUGrid *ugrid, PetscInt v_id1, PetscInt v_id2, PetscInt v_id3, PetscReal *area) {
```
jeff-cohere, line 972: Consider renaming this to `ComputeTriArea` or something that indicates that we're computing the area of a triangle.

src/fv/share/tdymeshcustom.c
```diff
@@ -0,0 +1,1315 @@
...
+/// @param [in] v_id3 Third vertex ID of the triangel
+/// @param [in] point_up Coordinate of upwind cell
+/// @param [in] point_dn Coordinate of downwind cell
+/// @param [in] n_up_dn Unit vector joining upwind and downwind cells
+/// @param [out] area Area of the triange
+/// @param [out] area_projected Projected area along unit normal
+/// @param [out] intercept Intercept between upwind-downind line to the triangle
+///
+/// @returns 0 on success, or a non-zero error code on failure
+static PetscErrorCode ComputeAreasAndIntercpet(TDyUGrid *ugrid, PetscInt v_id1, PetscInt v_id2, PetscInt v_id3, PetscReal point_up[3],
```
jeff-cohere, line 1014: Consider `ComputeTriAreasAndIntercept` (and watch out for the `Intercpet` typo)

src/fv/share/tdymeshcustom.c
```diff
@@ -0,0 +1,1315 @@
...
+
+/// Computes following geometric attributes of all faces in the TDyUGrid struct
+///  - centroid
+///  - area
+///
+/// @param [in] cell_to_face Mapping of ith-face of a cell to face id
+/// @param [inout] ugrid A TDyUGrid struct
+///
+/// @returns 0 on success, or a non-zero error code on failure
+static PetscErrorCode ComputeGoeAttrOfUGridFaces(PetscInt **cell_to_face, TDyUGrid *ugrid) {
```
jeff-cohere, line 1162: `GoeAttr` -> `GeoAttr`

jeff-cohere (changes requested):
Okay, I've made a first complete pass. There's too much here to review everything closely. This PR moves us toward the goal of showing something at the ESMD PI meeting, which is great. Probably some of this code will only stick around temporarily--we can factor out commonalities whenever it's appropriate.

Nevertheless, I found some things that we should look at more closely.

src/tdycore.c
bishtgautam: The way the TDycore code (outside of this PR) is implemented, we have to use `ierr = tdy->ops->time_cut(tdy);` instead of `ierr = tdy->ops->time_cut(&tdy->context, ...);` Would `ierr = tdy->ops->time_cut(tdy);` be a valid polymorphic function?

src/fv/share/tdymeshcustom.c
bishtgautam: Fixed as suggested.

src/fv/share/tdymeshcustom.c
bishtgautam: Fixed as suggested.

src/fv/share/tdyugrid.c
bishtgautam: Fixed.

src/fv/share/tdyugrid.c
bishtgautam: Fixed.

src/tdycore.c
bishtgautam: Fixed.

include/private/tdydiscretizationimpl.h
bishtgautam: Fixed.

src/fv/share/tdyugrid.c
bishtgautam: I used `char filename[PETSC_MAX_PATH_LEN]`

src/tdycore.c
jeff-cohere: Good point. I guess we haven't really talked about this much. When I refactored TDycore to be polymorphic, I wanted to make it clear which information was available to each function instead of just passing the `TDy`object itself, which has pretty much everything. This is an aesthetic design choice, but it can be helpful to "suggest" which data a function "should" or "shouldn't" modify.

In this case, maybe we can just pass the `TDy` object as you have done. If we later find a pattern to our approaches for cutting the timestep, we can always go back and change the function to take specific data instead of `TDy`.

jeff-cohere:
I rebased this branch against `master` to fix a merge conflict. @bishtgautam , I agree that the memory inconsistencies can be sorted in a separate smaller PR. Let's make sure the tests pass and then I can merge (unless @jedbrown wants to weigh in).

PR #231, opened 2022-06-02
Fixed a few floating point literals.

jeff-cohere:
In C, floating point literal values can be expressed many ways, but if
one uses scientific notation, `e` should be used instead of `d` as the
exponent indicator.


bishtgautam:
I guess I was in the Fortran world when I wrote the code.

PR #240, opened 2022-08-04
Commit-by-commit port of Rosie's salinity PR to new code structure

jeff-cohere:
Hi everyone,

@leorosie and I were working on a lot of things at the same time, so our paths diverged. This PR is a reimagining of PR #239 in the new code structure in which material models, characteristic curves, etc are factored out, and our new "polymorphic" model is in place, with vectorized functions and all that wacky stuff.

I've revised each of Rosie's commits and stuck both of our names on them using a convention suggested by @jedbrown (thanks!). Hopefully the code is similar enough for us to evaluate it alongside PR #239 . I took the liberty of changing the names of a few things ("concentration" to "salinity" to be more specific, and "PorDiff" to "SalineDiffusivity", etc), but these are just suggestions.

It looks like there may be some test failures in the transient MPFAO F90 demos. Perhaps they result from the slight change in the "constant" density of water introduced in this PR.

### DIFF
--- a/src/fv/mpfao/tdympfao.c
+++ b/src/fv/mpfao/tdympfao.c
@@ -320,17 +320,15 @@ static PetscErrorCode AllocateMemoryForBoundaryValues(TDyMPFAO *mpfao,
   }
 
   PetscInt i;
-  PetscReal dden_dP, dden_dPsi, d2den_dP2, dmu_dP, dmu_dPsi, d2mu_dP2;
+  PetscReal dden_dP, d2den_dP2, dmu_dP, d2mu_dP2;
   for (i=0;i<nbnd_faces;i++) {
     PetscReal m_nacl = mpfao->m_nacl[0];
-    PetscReal dm_nacl = mpfao->dm_nacl[0];
-    PetscReal d2m_nacl = mpfao->d2m_nacl[0];
     ierr = EOSComputeWaterDensity(eos,
-      mpfao->Pref, mpfao->Tref, m_nacl, dm_nacl, d2m_nacl,
-      &(mpfao->rho_bnd[i]), &dden_dP, &dden_dPsi, &d2den_dP2); CHKERRQ(ierr);
+      mpfao->Pref, mpfao->Tref, m_nacl,
+      &(mpfao->rho_bnd[i]), &dden_dP, &d2den_dP2); CHKERRQ(ierr);
     ierr = EOSComputeWaterViscosity(eos,
       mpfao->Pref, mpfao->Tref, m_nacl,
-      &(mpfao->vis_bnd[i]), &dmu_dP, &dmu_dPsi, &d2mu_dP2); CHKERRQ(ierr);
+      &(mpfao->vis_bnd[i]), &dmu_dP, &d2mu_dP2); CHKERRQ(ierr);
   }
 
   TDY_STOP_FUNCTION_TIMER()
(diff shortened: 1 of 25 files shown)


src/fv/mpfao/tdympfao.c
```diff
@@ -794,6 +862,15 @@ PetscErrorCode TDySetDMFields_TH_MPFAO(void *context, DM dm) {
   PetscFunctionReturn(0);
 }
 
+PetscErrorCode TDySetDMFields_Salinity_MPFAO(void *context, DM dm) {
+  PetscFunctionBegin;
+  PetscErrorCode ierr;
+  // Set up the section, 2 dofs per cell.
+  ierr = SetFields(dm, 2, (const char*[2]){"LiquidPressure", "Concentration"},
```
bishtgautam, line 869: `Concentration` --> `SalineConcentration`

src/materials/tdyeos.c
```diff
@@ -6,7 +6,7 @@ static PetscErrorCode ComputeWaterDensity_Constant(PetscReal p, PetscReal *den,
 
   PetscFunctionBegin;
 
-  *den = 998.0;
+  *den = 997.205;
```
bishtgautam, line 9: This will cause a lot of differences in testing. My guess is this change was made by @leorosie to match results for the new salinity model with PFLOTRAN, but we should revert this change.

src/materials/tdymaterialproperties.c
```diff
@@ -443,289 +530,233 @@ static PetscErrorCode FullTensorWrapperFunction(void *context, PetscInt n,
...
-  PetscFunctionBegin;
-
-  void *context;
-  PetscReal value[9];
-  value[0] = porosity;
-  ierr = CreateConstantContext(matprop->dim, value, &context); CHKERRQ(ierr);
-  ierr = MaterialPropSetPorosity(matprop, context, ConstantScalarWrapperFunction,
-                                 TDyFree); CHKERRQ(ierr);
-  PetscFunctionReturn(0);
+// These macros are used to support the various material model quantities.
```
bishtgautam, line 533: This is nice. I never think about using macros and this provides a concrete example of when one should consider using a macro.

src/fv/mpfao/tdympfao.c
```diff
@@ -408,7 +436,7 @@ PetscErrorCode TDyCreate_MPFAO(void **context) {
 
   // Initialize defaults and data.
   mpfao->gmatrix_method = MPFAO_GMATRIX_DEFAULT;
-  mpfao->bc_type = DIRICHLET_BC;
+  mpfao->bc_type = NEUMANN_BC;
```
bishtgautam, line 439: I think this change in the default setting might also break a lot of tests.

bishtgautam (commented):
Thanks for @jeff-cohere for getting this rebased. I have a few minor comments and addressing some of those comments would require input from @leorosie.

src/fv/mpfao/tdympfao.c
jeff-cohere: I should also change `SetBoundarySalinity` (and maybe the corresponding source/sink term) to match whatever term we select.

jeff-cohere:
Hey @leorosie ,

I don't know if you're back in front of your glowing screen, but if so, I hope you had a nice break!

Do you mind if I change the water density value and the boundary condition default option that @bishtgautam has mentioned above? And do you have thoughts on this PR overall (names, organization, etc)?

src/materials/tdyeos.c
leorosie: Yes that's correct, you can revert it!

src/fv/mpfao/tdympfao.c
leorosie: This change was also made to try and match results with PFLOTRAN and can be reverted.

leorosie:
@jeff-cohere The water density value and boundary condition default option were changed to match PFLOTRAN so those can both be changed back. The rest looks good to me, thanks for taking this on!

jeff-cohere:
Okay, I think all this is ready. We don't have regression tests yet for salinity, though. @bishtgautam , do you want those to be added in a subsequent PR, or should we add them to this one before we merge?

bishtgautam:
@jeff-cohere Let's add a new test in the PR itself. Thanks.

jeff-cohere:
> @jeff-cohere Let's add a new test in the PR itself. Thanks.

@leorosie , are you comfortable coming up with such a test (or borrowing one from PFLOTRAN)? I have a cartoon-level understanding of the equations, but I am no salinographer. :-)

bishtgautam (approved):
We can merge this PR now and will add a new regression test in a separate PR.

PR #243, opened 2022-09-08
Update PetscOptionsBegin/End for petsc-3.18/main

jedbrown:
We could consider adopting `PetscCall` notation and clang-format if others like it, but this gets things compiling (and fixes a couple format strings).


jeff-cohere (approved):
Thanks, @jedbrown . Do you think we should move to PETSc 3.18 so we don't break compiles?

jeff-cohere:
Also, I'm all for `clang-format` (though I have noticed that its output doesn't always seem stable between versions).

jedbrown:
I think this will all compile with older PETSc too.

Yeah, clang-format changes a little between versions, but it's nice to remove the cognitive load of anything to do with formatting.

